PR #5, opened 2016-10-25
Add CI configuration

kriskowal:
r @prashantv 

I think this is also sufficient for coveralls. We should enable that for this repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+sudo: false
+language: go
+
+go:
+    - 1.7
+
+env:
+    global:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,21 @@
+sudo: false
+language: go
+
+go:
+    - 1.7
+
+env:
+    global:
+        - GO15VENDOREXPERIMENT=1
```
prashantv: since we only use 1.7 we can drop this (it's enabled by default in 1.6 and always on in 1.7)

rabbbit:
Will be done in #35

PR #8, opened 2016-12-10
Get rid of nap with gopeek

cat2neat:
The below is quoted from https://github.com/uber-go/ratelimit/pull/1#discussion-diff-78767601 .

> Alternatively, if this is purely for testing, we could have a "sleep till there's no runnable goroutines" function. There's no easy way to do it, but we'd get a list of all goroutines, and make sure that they're all blocked on something.

As I felt there are not negligible such needs especially for testing
I've developed https://github.com/cat2neat/gopeek to realise the above requirements in a generalized way.

This PR would
- get rid of nap with gopeek
- get the test time speed up around 3times.
- get the test more stable in a resource constrained one like travis-ci as time.Sleep tend to be unstable

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -5,3 +5,4 @@ testImport:
   subpackages:
   - assert
 - package: github.com/uber-go/atomic
+- package: github.com/cat2neat/gopeek
--- a/internal/clock/clock.go
+++ b/internal/clock/clock.go
@@ -24,8 +24,12 @@ package clock
 
 import (
 	"container/heap"
+	"fmt"
+	"os"
 	"sync"
 	"time"
+
+	"github.com/cat2neat/gopeek"
 )
 
 // Mock represents a mock clock that only moves forward programmically.
@@ -44,12 +48,19 @@ func NewMock() *Mock {
 
 // Add moves the current time of the mock clock forward by the duration.
 // This should only be called from a single goroutine at a time.
-func (m *Mock) Add(d time.Duration) {
+func (m *Mock) Add(d time.Duration, condition *gopeek.Condition) {
 	m.Lock()
 	// Calculate the final time.
 	end := m.now.Add(d)
 
 	for len(m.timers) > 0 && m.now.Before(end) {
+		// Wait before popping from timers
+		m.Unlock()
+		_, err := condition.Wait(time.Second)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error: %+v\n", err)
+		}
+		m.Lock()
 		t := heap.Pop(&m.timers).(*Timer)
 		m.now = t.next
 		m.Unlock()
@@ -58,8 +69,6 @@ func (m *Mock) Add(d time.Duration) {
 	}
 
 	m.Unlock()
-	// Give a small buffer to make sure the other goroutines get handled.
-	nap()
 }
 
 // Timer produces a timer that will emit a time some duration after now.
@@ -94,7 +103,6 @@ func (m *Mock) AfterFunc(d time.Duration, f func()) *Timer {
 		<-t.c
 		f()
 	}()
-	nap()
 	return t
 }
 
@@ -126,8 +134,4 @@ func (t *Timer) Tick() {
 	case t.c <- t.next:
 	default:
 	}
-	nap()
 }
-
-// Sleep momentarily so that other goroutines can process.
-func nap() { time.Sleep(1 * time.Millisecond) }
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -9,6 +9,7 @@ import (
 	"go.uber.org/ratelimit"
 	"go.uber.org/ratelimit/internal/clock"
 
+	"github.com/cat2neat/gopeek"
 	"github.com/stretchr/testify/assert"
 	"github.com/uber-go/atomic"
 )
@@ -79,9 +80,14 @@ func TestRateLimiter(t *testing.T) {
 		wg.Done()
 	})
 
-	clock.Add(4 * time.Second)
+	condition := gopeek.NewCondition().
+		CreatedBy("ratelimit_test.TestRateLimiter").
+		Is(gopeek.StateWaitingChannel).
+		// go ahead when there is only one waiting channel via Mock.Sleep
+		// as it's evidence that ratelimit occurred
+		EQ(1)
 
-	clock.Add(5 * time.Second)
+	clock.Add(4*time.Second, condition)
 }
 
 func TestDelayedRateLimiter(t *testing.T) {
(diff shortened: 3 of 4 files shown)


glide.yaml
```diff
@@ -5,3 +5,4 @@ testImport:
   subpackages:
   - assert
 - package: github.com/uber-go/atomic
+- package: github.com/cat2neat/gopeek
```
prashantv: Can you also do `glide up` so that `glide.lock` is also updated

internal/clock/clock.go
```diff
@@ -44,12 +48,19 @@ func NewMock() *Mock {
...
 // Add moves the current time of the mock clock forward by the duration.
 // This should only be called from a single goroutine at a time.
-func (m *Mock) Add(d time.Duration) {
+func (m *Mock) Add(d time.Duration, condition *gopeek.Condition) {
 	m.Lock()
 	// Calculate the final time.
 	end := m.now.Add(d)
 
 	for len(m.timers) > 0 && m.now.Before(end) {
+		// Wait before popping from timers
```
prashantv: Can you explain why we need to wait before popping timers? I would expect a wait after popping a timer.

ratelimit_test.go
```diff
@@ -123,11 +129,18 @@ func TestDelayedRateLimiter(t *testing.T) {
...
 		wg.Done()
 	})
 
-	clock.Add(40 * time.Second)
+	condition := gopeek.NewCondition().
+		CreatedBy("ratelimit_test.TestDelayedRateLimiter").
+		Is(gopeek.StateWaitingChannel).
+		// go ahead when there is >= 1 waiting channel via Mock.Sleep
+		// as it's evidence that ratelimit occurred
+		GT(0)
```
prashantv: Slight inconsistency here: comment says >= 1, but code says `GT(0)`

cat2neat:
Thank you for the FB @prashantv .

> The simple cases are fine, but what happens when I want to wait until there's 2 goroutines waiting for a channel, and one goroutine waiting on a lock?
> At the same time, I do appreciate the convenience (which is why we use testify), I just want to make sure the API allows for more flexibility if required.

like this. 

```go
gopeek.NewCondition().
       CreatedBy("gopeek_test.TestGoPeek.*").
       EQ(2).
       FilterByGoes(func(gs []stack.Gorotine) bool {
           for _, g := range gs {
               // `${Condition}` means any exported fields in [stack.Gorotine](https://github.com/maruel/panicparse/blob/master/stack/stack.go) is used to identify a specific gorotine
               if g.${Condition} && gopeek.NewState(g.State) == gopeek.StateWaitingChannel {
                   continue
               } else if gopeek.NewState(g.State) == gopeek.StateWaitingLock {
                   continue
               }
               return false
           }
           return true
       }).Wait(time.Second)
```

The complex cases typically can be solved using user-defined filter (FilterByGoes or FilterByGo) with stack.Gorotine .

glide.yaml
cat2neat: done.

internal/clock/clock.go
cat2neat: Without nap ( Strictly speaking the current test can fail even with nap if a gorotine invoking time.Sleep wake up before workers start to run. actually I've faced sometimes while testing ) 
there is a possibility that 
```go
t := heap.Pop(&m.timers).(*Timer)
```
may be invoked before one of workers running job get into the point [clock.Sleep](https://github.com/uber-go/ratelimit/blob/master/ratelimit.go#L121) .
in this case, as there are still only timers registered by clock.AfterFunc , test will fail.

cat2neat:
@kriskowal @prashantv 
It turned out that this PR may cause a little bit drawback especially under less vcpu on linux.
ExampleRatelimit get more flaky (though already flaky as is)
because gopeek use runtime.Gosched with a tight loop instead of nap(time.Sleep) while waiting for getting goroutines satisfy a specified condition that means ratelimit.test get more cpu intensive
so the linux default scheduler [CFS](https://en.wikipedia.org/wiki/Completely_Fair_Scheduler) tend to be late giving back a timeslice to ratelimit.test cause time.Sleep unstable like the following.

```shell
--- PASS: ExampleRatelimit (0.09s)
PASS
ok      go.uber.org/ratelimit   1.998s
=== RUN   TestUnlimited
--- PASS: TestUnlimited (0.00s)
=== RUN   TestRateLimiter
--- PASS: TestRateLimiter (0.40s)
=== RUN   TestDelayedRateLimiter
--- PASS: TestDelayedRateLimiter (1.59s)
=== RUN   ExampleRatelimit
--- PASS: ExampleRatelimit (0.09s)
PASS
ok      go.uber.org/ratelimit   2.088s
=== RUN   TestUnlimited
--- PASS: TestUnlimited (0.00s)
=== RUN   TestRateLimiter
--- PASS: TestRateLimiter (0.22s)
=== RUN   TestDelayedRateLimiter
--- PASS: TestDelayedRateLimiter (1.45s)
=== RUN   ExampleRatelimit
--- PASS: ExampleRatelimit (0.09s)
PASS
ok      go.uber.org/ratelimit   1.785s
=== RUN   TestUnlimited
--- PASS: TestUnlimited (0.00s)
=== RUN   TestRateLimiter
--- PASS: TestRateLimiter (0.32s)
=== RUN   TestDelayedRateLimiter
--- PASS: TestDelayedRateLimiter (1.63s)
=== RUN   ExampleRatelimit
--- PASS: ExampleRatelimit (0.10s)
PASS
ok      go.uber.org/ratelimit   2.048s
=== RUN   TestUnlimited
--- PASS: TestUnlimited (0.00s)
=== RUN   TestRateLimiter
--- PASS: TestRateLimiter (0.32s)
=== RUN   TestDelayedRateLimiter
--- PASS: TestDelayedRateLimiter (1.57s)
=== RUN   ExampleRatelimit
--- PASS: ExampleRatelimit (0.09s)
PASS
ok      go.uber.org/ratelimit   1.990s
=== RUN   TestUnlimited
--- PASS: TestUnlimited (0.00s)
=== RUN   TestRateLimiter
--- PASS: TestRateLimiter (0.23s)
=== RUN   TestDelayedRateLimiter
--- PASS: TestDelayedRateLimiter (1.65s)
=== RUN   ExampleRatelimit
--- FAIL: ExampleRatelimit (0.10s)
got:
1 10ms
2 10.514905ms
3 9.485095ms
4 10ms
5 10ms
6 10ms
7 10ms
8 10ms
9 10ms
want:
1 10ms
2 10ms
3 10ms
4 10ms
5 10ms
6 10ms
7 10ms
8 10ms
9 10ms
FAIL
exit status 1
FAIL    go.uber.org/ratelimit   1.988s
Error occurred at 6th test.
```

Workarounds I come up with are
1. add sufficient time.Sleep to the bottom of TestDelayedRateLimiter
or
2. replace ratelimit.New(100) with ratelimit.New(10) // reduce the rate
or
3. delete "Output:" in ExampleRatelimit

IMHO,
As ExampleRatelimit is already flaky even without this PR,
2 (reduce the rate) looks promising one to me.
Thoughts?

rabbbit:
Hi,

I don't think we'll be merging this (after 3 years;)).

We've made some changes to our test runners and we no longer need to `time.Sleep()`.

Issue #10, opened 2017-10-04
Parity with yab ratelimit


achals:
I was trawling through the yab ratelimiting code, and noticed that the [interface exposed](https://github.com/yarpc/yab/blob/dev/ratelimit/time_period.go#L35) there is different (and IMO nicer) than the interface in this package.

Is this package intended to be kept updated? If so, would it be possible to update the implementation to bring it up the parity? I'd be more than happy to contribute a PR if it would help. It would be great to be able to use rate limiting logic independently of yab.

kriskowal:
A change of this kind would require a major version bump. It could be done, since nothing we know of uses this package yet. Ideally we *would* use this package as a dependency in yab. I’d like to know what @prashantv thinks.

kriskowal:
To clarify, the difference is that yab’s take accepts a cancel channel.

prashantv:
Since we haven't cut any releases, I think it's fine to change the interface.

I think the ideal interface would take a `context.Context` for cancellation.

achals:
So I have a branch locally that updates the interface to accept a `context.Context`; it seems to be mostly working but `ExampleRateLimit` has way more variance than earlier:

```
--- FAIL: ExampleRatelimit (0.09s)
got:
1 10.935692ms
2 9.814699ms
3 11.718343ms
4 9.582154ms
5 8.356239ms
6 12.062452ms
7 8.729969ms
8 8.911591ms
9 9.937449ms
want:
1 10ms
2 10ms
3 10ms
4 10ms
5 10ms
6 10ms
7 10ms
8 10ms
9 10ms
FAIL
```

Again, go newb, but is switching to using `select` to block on channels supposed to add that kind of over head? I can make a PR if it helps the discussion.

achals:
Ignore my earlier comment, I was misusing the `Clock` interface, which I've actually removed in my PR.

nathanhack:
Is there a status on when the PR for this might be merged?

PR #11, opened 2017-10-09
Update the Limiter interface to accept a context.Context object

achals:
This also removes the Clock abstraction, since there doesn't seem to be a way to create timers using the mock clock effectively, and it isn't used in yab at all. I can add it back in and try to munge it to work if it has value.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -25,6 +25,8 @@ import (
 	"time"
 
 	"go.uber.org/ratelimit/internal/clock"
+	"context"
+	"math"
 )
 
 // Note: This file is inspired by:
@@ -35,7 +37,7 @@ import (
 // may block to throttle the goroutine.
 type Limiter interface {
 	// Take should block to make sure that the RPS is met.
-	Take() time.Time
+	Take(context.Context) bool
 }
 
 // Clock is the minimum necessary interface to instantiate a rate limiter with
@@ -49,6 +51,7 @@ type Clock interface {
 type limiter struct {
 	sync.Mutex
 	last       time.Time
+	timer      *time.Timer
 	sleepFor   time.Duration
 	perRequest time.Duration
 	maxSlack   time.Duration
@@ -63,6 +66,7 @@ func New(rate int, opts ...Option) Limiter {
 	l := &limiter{
 		perRequest: time.Second / time.Duration(rate),
 		maxSlack:   -10 * time.Second / time.Duration(rate),
+		timer:      time.NewTimer(time.Duration(math.MaxInt64)),
 	}
 	for _, opt := range opts {
 		opt(l)
@@ -91,7 +95,7 @@ func withoutSlackOption(l *limiter) {
 
 // Take blocks to ensure that the time spent between multiple
 // Take calls is on average time.Second/rate.
-func (t *limiter) Take() time.Time {
+func (t *limiter) Take(ctx context.Context) bool {
 	t.Lock()
 	defer t.Unlock()
 
@@ -100,7 +104,7 @@ func (t *limiter) Take() time.Time {
 	// If this is our first request, then we allow it.
 	if t.last.IsZero() {
 		t.last = now
-		return t.last
+		return true
 	}
 
 	// sleepFor calculates how much time we should sleep based on
@@ -108,6 +112,7 @@ func (t *limiter) Take() time.Time {
 	// Since the request may take longer than the budget, this number
 	// can get negative, and is summed across requests.
 	t.sleepFor += t.perRequest - now.Sub(t.last)
+	t.last = now
 
 	// We shouldn't allow sleepFor to get too negative, since it would mean that
 	// a service that slowed down a lot for a short period of time would get
@@ -118,14 +123,18 @@ func (t *limiter) Take() time.Time {
 
 	// If sleepFor is positive, then we should sleep now.
 	if t.sleepFor > 0 {
-		t.clock.Sleep(t.sleepFor)
+		t.timer.Reset(t.sleepFor)
+		select {
+		case <-t.timer.C:
+		case <-ctx.Done():
+			return false
+		}
+
 		t.last = now.Add(t.sleepFor)
 		t.sleepFor = 0
-	} else {
-		t.last = now
 	}
 
-	return t.last
+	return true
 }
 
 type unlimited struct{}
@@ -135,6 +144,6 @@ func NewUnlimited() Limiter {
 	return unlimited{}
 }
 
-func (unlimited) Take() time.Time {
-	return time.Now()
+func (unlimited) Take(_ context.Context) bool{
+	return true
 }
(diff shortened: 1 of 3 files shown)


achals:
Addresses #10

achals:
cc @kriskowal @prashantv

ratelimit.go
```diff
@@ -24,7 +24,8 @@ import (
 	"sync"
 	"time"
 
-	"go.uber.org/ratelimit/internal/clock"
+	"context"
```
prashantv: nit: no blank between stdlib imports

ratelimit.go
```diff
@@ -63,24 +56,14 @@ func New(rate int, opts ...Option) Limiter {
 	l := &limiter{
 		perRequest: time.Second / time.Duration(rate),
 		maxSlack:   -10 * time.Second / time.Duration(rate),
+		timer:      time.NewTimer(time.Duration(math.MaxInt64)),
```
prashantv: rather than use a max duration, should we create a timer, stop it (maybe in `init`) and use that everywhere?

ratelimit.go
```diff
@@ -35,24 +36,16 @@ import (
...
-	Take() time.Time
-}
-
-// Clock is the minimum necessary interface to instantiate a rate limiter with
-// a clock or mock clock, compatible with clocks created using
-// github.com/andres-erbsen/clock.
-type Clock interface {
-	Now() time.Time
-	Sleep(time.Duration)
+	Take(context.Context) bool
```
prashantv: what does the `bool` argument mean?

ratelimit.go
```diff
@@ -118,14 +102,19 @@ func (t *limiter) Take() time.Time {
 
 	// If sleepFor is positive, then we should sleep now.
 	if t.sleepFor > 0 {
-		t.clock.Sleep(t.sleepFor)
+		t.timer.Stop()
+		t.timer.Reset(t.sleepFor)
```
prashantv: stopping and resetting a timer is racy, take a look at the documentation for how to stop and reset:
https://golang.org/pkg/time/#Timer.Reset
```
if !t.Stop() {
	<-t.C
}
t.Reset(d)
```

prashantv (changes requested):
Overall, I like the direction of this change, but we also haven't motivated the reasoning for some of these changes

 * The interface is now a blocking interface (similar to `yab`) rather than letting the caller do the sleep. It makes sense, but I'd like some motivation for the change so we don't just flip-flop between the 2 interfaces
 * Since we don't use a mock clock, tests now rely on real time (both within this package, and external packages using it). We need a way of testing that doesn't end up sleeping for some number of real seconds. Tests that rely on real time tend to be flaky, or end up taking a long time.

ratelimit.go
achals: Done

ratelimit.go
achals: Right now, `Take` returns a `True` if it was unblocked without any cancellations from the passed in context, and `False` if it was unblocked because of the context.

ratelimit.go
achals: I can also leave the `timer` field  as `nil` within the struct and have a check for it in `Take` & set it when needed - but I'm not sure if there are any performance implications.

ratelimit.go
achals: Good catch, thanks!

achals:
@prashantv Thanks a lot for the feedback! Within Hailstorm, and also generally speaking, it seems like the `Limiter` interface is meant to be used explicitly in the case when some operation is meant to be repeated and maintain a specific RPS - i.e. there would probably need to be blocking in the most common use cases. 

In that case, having the goroutine block as a result of the call to the `Limiter` seems like the most complete experience, from a users' perspective, and frees me from having to rebuild blocking code which might be error prone.

For testing, I'm certainly open to adding back the mock clock since I like the abstraction as well - lemme see if there's a way of accomplishing that.

achals:
Also @prashantv , correct me if I'm wrong but the old interface still blocking, is it not? The docs as well as the implementation state that the Limiter is meant to sleep till the rps is met.

ratelimit.go
prashantv: still see a blank between "time" and "context"

ratelimit.go
prashantv: Can we update the documentation to make that part of the API

I'm also not sure how valuable it is, since the caller can just check `ctx.Err() != nil`

What is the value in having this in the API?

ratelimit.go
prashantv: `nil` check works too, but less branches is always nice, so might be better to have an expired timer.

prashantv (commented):
So this package originally was a copy of yab's rate limiter, which was blocking. However, it was then modified to be non-blocking. This is why I wanted to make sure we had good motivation for making it blocking again.

For example, we have use cases where we don't want to sleep, but want to deny the request if it exceeds the rate limit. Ideally the same code could be used for both.

jeromefroe:
Is there any interest in reviving this PR? I have a use case where I would like to be able to pass the `Limiter` a `Context` so that it can return early if the `Context` is cancelled. Currently, the code looks like the following:

```go
type Thing struct {
    limiter ratelimit.Limiter
}

func (t *Thing) Do(ctx context.Context) error {
    t.limiter.Take()

    select {
    case <-ctx.Done():
         return ctx.Err()
    default:
    }

    // expensive work ...
}
```

The code works as is, but it would be nice, and more efficient, if I could combine the two steps into one:

```go
func (t *Thing) Do(ctx context.Context) error {
    if _, err := <-t.limiter.TakeWithCtx(ctx); err != nil {
        return err
     }

    // expensive work ...
}
```

Implementation-wise it should, in theory, be relatively trivial to support such a method. The heart of `TakeWithCtx` could look like the following:

```go
func (t *limiter) TakeWithCtx() (time.Time, error) {
    // setup work to determine the amount of time we should sleep for ...

    timer := time.NewTimer(t.sleepFor)
    select {
    case <-ctx.Done():
        // possible cleanup work ...

        timer.Stop()
        return time.Time{}, ctx.Error()
    case <-timer.C:
    }

    // ...
}
```

In practice I imagine this change might be more difficult though because:

1. It requires a new method, `TakeWithCtx`, on the `Limiter` interface which will be a breaking change.
2. It requires the `Clock` interface to be re-done as `TakeWithCtx` won't work if the `Sleep` method is blocking, another breaking change.

I'd be curious to know if anyone else see values in such a change?

prashantv:
This repo has no releases so breaking changes are fine. I don't think there's really an owner right now -- this is a copy of the rate limiting code in yab which I wrote for limiting the number of requests sent by yab. However, this library isn't used by yab, we can iterate on the APIs to find something that makes sense for everyone.

achals:
I'm up for reviving this PR, if it will prove useful. I'll spend some time time addressing @prashantv's comments from last time and update.

PR #12, opened 2017-10-20
Add a Reset method to the Limiter interface

achals:
We'd love to be able to Reset the rps on the fly without creating a new Limiter entirely.


prashantv:
This will complicate the implementations, so I want to make sure we really need this before extending the interface.

What is the use case where `Reset` with a new `RPS` is used frequently enough that creating a new `Limiter` would impact performance?

PR #15, opened 2018-10-23
Atomic based limiter

storozhukBM:
Hi,
I've made your limiter implementation atomic based. It is a little bit harder but in generally faster and has no goroutine scalability issues that any mutex based implementation have.
Checkout this relation between RateLimiter latency and count of goroutines:
limiter_scalability
Original implementation moved to mutexbased.go file, so you can run benchmarks and play with other tests. But I'd suggest to leave only one implementation after all.

[Side note] This wired latency spike in mutex based implementation is a direct illustration of mutex falling into [starvation mode](https://github.com/golang/go/blob/master/src/sync/mutex.go#L44).

Full benchmarking results [here](https://github.com/uber-go/ratelimit/files/2504561/full_test_with_procs_fit.pdf)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 test:
-	go test .
+	go test -race .
 cover:
 	go test . -coverprofile=cover.out
 	go tool cover -html=cover.out
+bench:
+	go test -bench=.
--- a/mutexbased.go
+++ b/mutexbased.go
@@ -16,7 +16,7 @@ type mutexLimiter struct {
 }
 
 // New returns a Limiter that will limit to the given RPS.
-func NewMutexBased(rate int, opts ...Option) Limiter {
+func newMutexBased(rate int, opts ...Option) Limiter {
 	l := &mutexLimiter{
 		perRequest: time.Second / time.Duration(rate),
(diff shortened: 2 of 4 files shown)


storozhukBM:
Hi @kriskowal,
Sorry for bothering you, but maybe you'll have some time to look at this changes.

mxplusb:
@kriskowal any thoughts here?

kriskowal:
This looks like a pretty fantastic change. Can you move the prior implementation into an internal package so we don’t grow the API surface?

Sorry for the extraordinary delay.

storozhukBM:
Hi @kriskowal 
There was no easy way to move `mutexLimiter` into `internal` package because it depends on some types from `ratelimit` package like `Clock`, `Limiter`, `Option`...
So I decided to make `newMutexBased` private, which is basically has the same results, and we don’t grow the API surface.

kriskowal:
Looks good. I’ll try to land, changelog, release this week.

peakle:
@storozhukBM where did you create this awesome benchmark visualization?

storozhukBM:
@peakle 
Exported benchmark results to csv and configured graph in Numbers(spreadsheets software that goes together with MacOS)

PR #16, opened 2018-12-21
change import package

t-xinlin:


prashantv:
Hi @t-xinlin, we intentionally use `go.uber.org` for our packages, which lets us move them around if necessary.

t-xinlin:
ok

PR #17, opened 2019-05-24
Add WithCustomDuration

Jleagle:
I was having issues setting the duration to higher than a second and came across this in a fork.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -89,8 +89,9 @@ func withoutSlackOption(l *limiter) {
 	l.maxSlack = 0
 }
 
-// WithCustomDuration is an option to set durations not based around seconds.
-func WithCustomDuration(amount int, duration time.Duration) Option {
+// WithCountPerDuration is an option to set durations not based around seconds.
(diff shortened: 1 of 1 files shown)


ratelimit.go
```diff
@@ -89,6 +89,13 @@ func withoutSlackOption(l *limiter) {
 	l.maxSlack = 0
 }
 
+// WithCustomDuration is an option to set durations not based around seconds.
+func WithCustomDuration(amount int, duration time.Duration) Option {
```
kriskowal: Oh, this is unfortunate, but clearly necessary. Can we name this `WithCountPerDuration` and clarify in the comment that this overrides the `rate` argument of the constructor?

kriskowal:
Thank you.

kriskowal:
It occurred to me in review that an even better name for this option is Per since it would read New(10, Per(time.Second)) for 10Hz. This requires threading the options a bit differently, so we can take advantage of the existing `rate` argument and make it less obvious that we made an API mistake.

#21

PR #19, opened 2019-07-29
Fix import path for atomic test dependency

### LINKED ISSUE #18: Incorrect import path for atomic
The *most correct* import path for `atomic` package should be `go.uber.org/atomic`, there is a mistake in the testing file:
https://github.com/uber-go/ratelimit/blob/c15da02342779cb6dc027fc95ee2277787698f36/ratelimit_test.go#L13

kriskowal:
Fixes #18

Adds a change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+### Fixed
+- Changed the import path for `go.uber.org/atomic` to its newer, canonical
+  import path.
+  [#18](https://github.com/uber-go/ratelimit/issues/18)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+### Fixed
+- Changed the import path for `go.uber.org/atomic` to its newer, canonical
+  import path.
+  [#18](https://github.com/uber-go/ratelimit/issues/18)
+
+## [1.0.0] - 2016-09-09
```
prashantv: don't think we ever released anything?

CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+- Initial release
+
+[Unreleased]: https://github.com/uber-go/fx/compare/v1.0.0...master
```
prashantv: wrong repo here and below

CHANGELOG.md
kriskowal: I’m tagging what stands on master so we can track.

kriskowal: backed this out. will follow with 0.1 tag.

PR #23, opened 2020-03-31
use the old state's sleepFor

Ventures-Adventures:
use the old state's sleepFor,  so the slack can be used for more than once


rabbbit:
Solved in #60.

PR #27, opened 2020-07-28
Need to get the last sleepFor in the loop.

QuantumEnergyE:
Need to get the last sleepFor in the loop, or slack will have no effect.


QuantumEnergyE:
TBR.

wanglongfei023:
cannot limit anymore as you code

rabbbit:
Solved in #60 - using your idea, but with tests/in a different file.

PR #28, opened 2020-08-04
Fix error use of internal package

noah8713:
When using ratelimit with go mod, it complaints of using internal module.
Reason is use of internal modules is not permitted from outside dep projects/packages.
e.g. ./../../../pkg/mod/github.com/uber-go/ratelimit@v0.1.0/ratelimit.go:27:2:
use of internal package go.uber.org/ratelimit/internal/clock not allowed

Hence, renaming the internal folder to central to avoid this failure using go mod.

Signed-off-by: Aliasgar Ginwala <aginwala@ebay.com>


prashantv:
Please import the package as `go.uber.org/ratelimit`, rather than `uber-go/ratelimit`, which should avoid this issue.

noah8713:
Thanks for the suggestion. I cleared my go.mod entries for uber-go and used go.uber.org/ratelimit. It works fine now.

PR #31, opened 2020-09-13
Switch from glide to Go Modules

rabbbit:
Per discussion not keeping glide for backwards compatibility,
go modules should be enough.

Also fixed a bad import in tests, came up during testing.


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Changed
+- Switched from Glide to Go Modules
```
abhinav, line 9:
```suggestion
- Switched from Glide to Go Modules.
```

go.mod
```diff
@@ -0,0 +1,8 @@
+module go.uber.org/ratelimit
+
+go 1.14
+
+require (
+	github.com/stretchr/testify v1.6.1
+	go.uber.org/atomic v1.6.0
```
abhinav, line 7: we just cut a new release. mind upgrading to 1.7.0?

CHANGELOG.md
rabbbit: <3 like this review (really do!) - will pay more attention.

go.mod
rabbbit: sure.

abhinav (approved):

PR #33, opened 2020-09-21
Add basic linting via a Makefile

rabbbit:
Largely based on https://github.com/uber-go/multierr/blob/master/Makefile.
Similarly to above using a submodule to avoid pulling dependencies.

Also fixed a few extra Makefile targets.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 ### Changed
 - Switched from Glide to Go Modules.
+- Basic linting setup via a Makefile.
```
abhinav, line 10: This doesn't need a changelog entry because it doesn't affect users in any way.

Makefile
```diff
@@ -1,7 +1,42 @@
...
+.PHONY: gofmt
+gofmt:
+	$(eval FMT_LOG := $(shell mktemp -t gofmt.XXXXX))
+	@gofmt -e -s -l $(GO_FILES) > $(FMT_LOG) || true
+	@[ ! -s "$(FMT_LOG)" ] || (echo "gofmt failed:" | cat - $(FMT_LOG) && false)
+
+.PHONY: golint
+golint:
+	@cd tools && go install golang.org/x/lint/golint
+	@$(GOBIN)/golint ./...
```
abhinav, line 30: consider breaking installation and execution into separate steps so that `make lint` doesn't always try to install golint.

```suggestion
bin/golint: tools/go.mod
	cd tools && go install golang.org/x/lint/golint

.PHONY: golint
golint: bin/golint
	bin/golint ./...
```

(golint binary has a dependency on tools/go.mod because that defines which version of golint we're using.)

Makefile
```diff
@@ -1,7 +1,42 @@
...
+
+.PHONY: lint
+lint: gofmt golint staticcheck
+
+.PHONY: staticcheck
+staticcheck:
+	@cd tools && go install honnef.co/go/tools/cmd/staticcheck
+	@$(GOBIN)/staticcheck ./...
```
abhinav, line 38: same as golint separation:


```suggestion
bin/staticcheck: tools/go.mod
	cd tools && go install honnef.co/go/tools/cmd/staticcheck

.PHONY: staticcheck
staticcheck: bin/staticcheck
	bin/staticcheck ./...
```

CHANGELOG.md
rabbbit: I think I've seen both approaches, but agreed, makes sense.

Makefile
rabbbit: Sure, make sense - didn't do it because I thought https://github.com/uber-go/multierr/blob/master/Makefile#L33 wasn't doing it for some smart reason :)

abhinav: No, just oversight

abhinav (approved):

abhinav:
Do you need to update the .travis.yml?

rabbbit:
> Do you need to update the .travis.yml?

Hm? There's no travis in this repo.

I intend to start looking at github actions soon.

PR #34, opened 2020-09-28
Fix linting issues.

rabbbit:
Now that we have linting(#33), we should make sure linting passes.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -55,6 +55,9 @@ type state struct {
 
 type limiter struct {
 	state unsafe.Pointer
+	//lint:ignore U1000 Padding is unused but it is crucial to maintain performance
+	// of this rate limiter in case of collocation with other frequently accessed memory.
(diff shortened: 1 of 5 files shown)


ratelimit.go
```diff
@@ -53,8 +54,7 @@ type state struct {
 }
 
 type limiter struct {
-	state   unsafe.Pointer
-	padding [56]byte // cache line size - state pointer size = 64 - 8; created to avoid false sharing
```
storozhukBM, line 57: You can't just remove padding, it is crucial to maintain performance properties of this rate limiter in case of collocation with other frequently accessed memory.

storozhukBM (changes requested):

ratelimit.go
rabbbit: Fair enough - I admit I didn't read/think this is important.

Added an explicit linter exception instead + copied over your comment.

cinchurge (approved):

PR #35, opened 2020-09-28
Configure CI via github actions

rabbbit:
Mostly based on https://github.com/uber-go/zap/commit/02a233f7b9041b29460a7372f788b324dbd64c4e

Planning to land and iterate? It seem to work fine, at least for the tests.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -44,6 +44,7 @@ jobs:
         run: make test
 
       - name: Upload coverage to codecov.io
+        if: matrix.latest
         uses: codecov/codecov-action@v1
 
       # TODO decide whether we want to benchmark on every run.
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,51 @@
+name: test
+
+on:
+  pull_request:
+    branches: ['*']
+
+jobs:
+  test:
+
+    name: Run tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        go: ["1.14.x", "1.15.x"]
+        include:
+        - go: 1.15.x
+          latest: true
+
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yaml
```diff
@@ -0,0 +1,59 @@
...
+        run: go mod download
+
+      - name: Lint
+        if: matrix.latest
+        run: make lint
+
+      - name: Test
+        run: make test
+
+      # TODO decide whether we want to benchmark on every run
```
rabbbit, line 49: This is somewhat slow currently - will probably have to skip, and manage manually.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,52 @@
+name: test
+
+on:
+  pull_request:
+    branches: ['*']
+
+jobs:
+  test:
+
```
cinchurge, line 9: are the blank lines necessary?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,52 @@
...
+    name: Run tests
+
+    runs-on: ubuntu-latest
```
cinchurge, line 12: is it possible to run this on latest debian as well?

rabbbit: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

No. Could run it on a mac, or windows, but I don't think it really matters? It's the same architecture anyway (?)

.github/workflows/test.yaml
rabbbit: They're not necessary, but it seems to be pattern people follow: https://docs.github.com/en/free-pro-team@latest/actions/quickstart and https://github.com/uber-go/zap/commit/02a233f7b9041b29460a7372f788b324dbd64c4e.

Could remove but it didn't seem horrible?

.github/workflows/test.yaml
cinchurge: yea it doesn't really matter i guess

cinchurge (approved):

PR #36, opened 2020-10-04
Fix `go vet` warning about bad Example naming

rabbbit:
See https://golang.org/pkg/testing/#hdr-Examples.

It could be `Example` or `ExampleLimiter` or `ExampleLimiter_Take`
but given we have a single Example example should be good enough.

Also removes an unnecessary `clock.Add` in a test.

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func ExampleRatelimit() {
+func Example() {
 	rl := ratelimit.New(100) // per second
 
 	prev := time.Now()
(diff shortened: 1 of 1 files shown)


ratelimit_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func ExampleRatelimit() {
+func Example() {
```
cinchurge, line 16: what should we do if we want to add more examples? e.g. with different implementations etc.

rabbbit: See commit message - we can do more thing by namespacing to struct/method, but also just appending suffixes, like `Example_One`, `Example_Two`.

abhinav: Yeah, you can rename this when you add more examples.

abhinav (approved):

Issue #3, opened 2016-09-28
Consolidate AfterFunc and Sleep in tests


kriskowal:
It should be sufficient to use Sleep in a goroutine and we should not need a separate AfterFunc.

Issue #22, opened 2019-12-13
Evaluate atomic and mutex based Take implementations.


twilly:
#15 brought in a new implementation of Take to avoid starving out the older mutex implementation (a major performance win). Before we cut a new release, let's dive a little deeper on the new implementation's behaviors and aim for a single implementation.

twilly:
The atomic implementation has two warts that make me weary. I don't like reaching for unsafe if it's at all avoidable. The fixed padding is also architecture specific. I'm not sure yet, but it may be possible  to avoid the unsafe pointer swap. I'm not too sure what to do about the padding.

The older mutex implementation held its lock during the blocking sleep call. Holding the lock doesn't appear to be necessary (compute and set last & sleepFor, unlock, sleep). Should this hold true, let's use the new benchmark to compare the three implementations.

rabbbit:
Nothing more planned.

PR #37, opened 2020-10-04
Move alternative implementations to internal/alternatives/*

rabbbit:
We currently have two different implementations. Currently:
- they're in the same package
- one of them is currently not tested
- one has a bug related to slack (see #27, #23)
- one has a broken constructor taking options affecting the other type

We plan to add at least 3 more implementations:
- one that addresses #22
- one that re-uses ratelimitfx atomic style
- one that compares us against x/time/rate

Even if we decide to eventually keep a single implementation,
they'll be there for a while. And I imagine it'll get noisy.

(A crazy idea would be to return different implementation based on
some user params. But that's totally TBD).

This diff tries to organize us a bit my moving the separate
implementations to separate packages.

Followups:
- will try to refactor the tests into a shared test package & re-use them.
  This way we can write tests once and make sure new implementations
  are working as expected.
- will try to make sure the implementations take equivalent
  configuration
- will have to modify the root ratelimit package to have stronger
  type guarantees, and not just aliasing methods - that would require
  a separate /types package though (I think).

Accidentally this diff is also changing signature of AfterFunc
in internal/clock package. This is so that it implements
it's own Clock interface. It's a bit of a temporary hack
as well - #3 will make it a bit simpler.

### DIFF
--- a/internal/alternatives/atomic/limiter.go
+++ b/internal/alternatives/atomic/limiter.go
@@ -1,0 +1,128 @@
+// Copyright (c) 2016, 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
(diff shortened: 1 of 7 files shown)


rabbbit:
About

> Accidentally this diff is also changing signature of AfterFunc
> in internal/clock package. This is so that it implements
> it's own Clock interface. It's a bit of a temporary hack
> as well - #3 will make it a bit simpler.

I tried to fix #3 in a separate PR, got stuck at something, will try again.
So we could potentially wait with landing this one.

But I'd still like to hear your thoughts on the idea:)

rabbbit:
Don't think we'll be doing this in the end.

PR #38, opened 2020-10-07
Prefactor types into a separarate types package

rabbbit:
This came up in #37 where we ended up in a dependency loop.
- main package was importing a specific rate limiter type
- both rate limiters needed to import Clock types from main package

By moving it to a separate package we avoid that loop.

If we decide to go ahead with #37, this will also be necessary
to have a "common config type" that can be re-used between all N
alternative limiters.

### DIFF
--- a/internal/types/types.go
+++ b/internal/types/types.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2016, 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package types
+
+import "time"
+
+// Limiter is used to rate-limit some process, possibly across goroutines.
+// The process is expected to call Take() before every iteration, which
+// may block to throttle the goroutine.
+type Limiter interface {
+	// Take should block to make sure that the RPS is met.
+	Take() time.Time
+}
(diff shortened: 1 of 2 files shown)


prashantv (commented):
This is going to make the API worse for users (I've seen this with zap / zapcore split as well).

Is there a big advantage in moving the implementations into separate packages vs having them unexported in the same package?

rabbbit:
> This is going to make the API worse for users (I've seen this with zap / zapcore split as well).
> 
> Is there a big advantage in moving the implementations into separate packages vs having them unexported in the same package?

In what way would it make it worse? I thought type aliasing would be invisible (?) - note we're still exporting only a single package unlike zap.

As to why:
- we currently have two implementations
- one has tests, one doesn't
- the tested one has a bug (lol)
- they have inconsistent options/constructors
-- and there are some feature requests to change the interface ("Per", allow cancellation, etc)
- the mutex one is accepting Option that belong to the other type
- **and**, we're planning to create/test at least 3 more implementations (x/rate, ratelimitfx, slightly different mutex one)

So I having a single interface/constructor options exported, and but splitting implementations would:
1. be significantly cleaner,
2. make it easier to re-use tests.

It wouldn't be impossible with a single package, just more messy.

I was also toying with a thought of returning different implementations based on user params - that would require more testing though. For now, the mutex implementation seems to be faster for a single goroutine though, but not sure if that will hold.

We could:
- land this, see if we actually need various implementations before 1.0, potentially revert to a single package then.
- do some tricks with implementations accepting `interface{}`, and type asserting to different clock type (we already have `ratelimit.Clock` and `ratelimit.internal.clock.Clock` interfaces). Would only need to happen in a single place, so it wouldn't be horrible.
- try to run with a single package

I'm for (1), but I guess it depends on how much worse it would

> make the API worse for users

What are you concerned about? :>

prashantv:
> In what way would it make it worse? I thought type aliasing would be invisible (?) - note we're still exporting only a single package unlike zap.

It's actually a little worse with type aliases to an internal package, since users will be forced to look at the API documentation for an internal package to understand the types fully. See how a type alias shows up in documentation:

https://pkg.go.dev/go.uber.org/zap#Field

There's basically no information, no method set, etc.

> It wouldn't be impossible with a single package, just more messy.

I understand the need for 2 implementations, the differences between them, but I don't think any of that explains why a separate package for each implementation helps. The implementations are small enough to be separate files, what advantages do we get from moving them into separate packages?

> land this, see if we actually need various implementations before 1.0, potentially revert to a single package then.

My preference is the other way around -- keep everything in a single package, and figure out workarounds for the dependency loop (which itself is actually caused by having multiple packages).

rabbbit:
> dependency loop which itself is actually caused by having multiple packages

Yeah, I understand what I'm causing;)

>  keep everything in a single package, and figure out workarounds for the dependency loop (which itself is actually caused by having multiple packages).

This I don't understand - if we keep everything in a single package, there's no loop, so there's no need for workarounds? ;)

> need for 2 implementations

Might get to 4-5, at least temporarily.

> what advantages do we get from moving them into separate packages?

The usual package benefits:
- separate namespace
-- no `NewMutexBased`, `NewAtomicBased`, `NewAtomicUnsafeBased`, `NewFake`, just `New`
-- no accidentally sharing structs, like `Option` being shared in wrong ratelimiters, as we have do right now
- separate `go test` and `go coverage`
- easier to enforce "every limiter should pass exactly same tests"
- I (maybe a bad pattern) find it easier to think about/separate: we have a common interface/tests, and we have N implementations that should share nothing. Packages mean "share nothing".

All that could still be achieved in a single package, but splitting it up "is free" - I don't have to think about it, go will check for me and make sure I make no silly mistakes.

rabbbit:
Spoke with @prashantv directly.

Agreed that type aliasing to a private package is not the way.

Package splitting it's still something I'd like to think about, but that can happen in #37.

PR #39, opened 2020-10-19
Get rid of internal/clock - Consolidate AfterFunc and Sleep in tests

### LINKED ISSUE #3: Consolidate AfterFunc and Sleep in tests

rabbbit:
Resolves #3

Was looking at the open issues and #3 seemed to make sense - it was
also something that I saw in #37 so I thought I'd take a stab.

The reasoning is that less code + smaller interface is a good thing.

During the diff it turned out that:
- our internal/clock has no tests
- our internal/clock has at least two bugs:
  - one that I'm fixing here, related to not executing two timers
   if they're scheduled at the "now" boundry
   - one regarding adding time but not having a timer set for that
   specific time (not a problem for us, so not touching it)

By dropping afterFunc into Sleep we're able to deprecate the internal
fork altogether - there is a missing nap in andres-erbsen/clock.AfterFunc
that meant we needed our own fork. Now it's not needed.


internal/clock/clock.go
```diff
@@ -49,7 +49,7 @@ func (m *Mock) Add(d time.Duration) {
 	// Calculate the final time.
 	end := m.now.Add(d)
 
-	for len(m.timers) > 0 && m.now.Before(end) {
+	for len(m.timers) > 0 && (m.now.Before(end) || m.now.Equal(end)) {
```
rabbbit, line 52: this is the bug I mentioned above.

internal/clock/clock.go
```diff
@@ -64,6 +64,8 @@ func (m *Mock) Add(d time.Duration) {
 
 // Timer produces a timer that will emit a time some duration after now.
 func (m *Mock) Timer(d time.Duration) *Timer {
+	m.Lock()
+	defer m.Unlock()
```
rabbbit, line 68: Need to move this from lines 79 to avoid a data race on m.now when creating timers.

rabbbit:
Rebased it on top of #40 and it looks much better now.

I'll raise another PR trying to switch us to andres-erbsen/clock in parallel to make sure we cannot simply do that, without even landing this diff.

rabbbit:
Raised #41 to test - verified that: 
- #41 on master fails
- #41 passes after #39 is merged - it looks like `AfterFunc` in andres-erbsen is indeed missing a nap.

internal/clock/clock.go
cinchurge: can we fix this but in its own diff? ideally it would contain only this change and an additional test that makes sure the boundary is accounted for

rabbbit: #41 drops our internal fork of the clock altogether.

I could raise a separate PR to fix the bug ... but I don't think adding tests (which currently don't exist at all :D) would make sense.

cinchurge:
change overall looks good, but i think we should split it up into 3 separate changes:
- the bug fix for the boundary check
- the fix for the race condition
- removal of AfterFunc along with updates to the test runner

what other work is needed in order to move off of the fork?

rabbbit:
> change overall looks good, but i think we should split it up into 3 separate changes:
> 
> * the bug fix for the boundary check
> * the fix for the race condition
> * removal of AfterFunc along with updates to the test runner
> 
> what other work is needed in order to move off of the fork?

I'm happy to split the fixes, but I won't be adding tests for them - since we're this diff away from removing the fork.

The work needed - if this lands, we can drop the fork.

internal/clock/clock.go
rabbbit: Okay, after a slack chat I merged #41 into this PR.

cinchurge (approved):
lgtm, thanks for making the changes!

PR #40, opened 2020-10-19
Add a test runner

rabbbit:
Something that came up in #39 - there's a fair bit of duplication
in the tests. A runner make current setup a bit easier/shorter,
but would make #39 much smaller.

There's no functional changes.

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -13,6 +13,38 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type runner struct {
+	wg     sync.WaitGroup
+	clock  *clock.Mock
+	count  atomic.Int32
+	doneCh chan struct{}
+}
+
+func runTest(t *testing.T, fn func(runner)) {
+	r := runner{
+		clock:  clock.NewMock(),
+		doneCh: make(chan struct{}),
+	}
+	defer close(r.doneCh)
+	defer r.wg.Wait()
+
+	fn(r)
+}
+
+func (r *runner) job(rl ratelimit.Limiter) {
+	go func() {
+		for {
+			rl.Take()
+			r.count.Inc()
+			select {
+			case <-r.doneCh:
+				return
+			default:
+			}
+		}
+	}()
+}
+
 func Example() {
 	rl := ratelimit.New(100) // per second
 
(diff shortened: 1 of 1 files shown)


ratelimit_test.go
```diff
@@ -47,95 +79,69 @@ func TestUnlimited(t *testing.T) {
...
+
+		r.clock.AfterFunc(1*time.Second, func() {
+			assert.InDelta(t, 100, r.count.Load(), 10, "count within rate limit")
+		})
+
+		r.clock.AfterFunc(2*time.Second, func() {
+			assert.InDelta(t, 200, r.count.Load(), 10, "count within rate limit")
+		})
+
+		r.wg.Add(1)
```
cinchurge, line 99: a test function being able to modify the state of the runner directly feels a little odd. would it be a better idea to pass in a runner interface instead? exposing the waitgroup directly allows for crazy things like calling wg.Wait inside the test function which can cause a deadlock. not that we expect people to do that, but it's better to reduce the API surface to avoid accidental problems

cinchurge:
I like the idea, but we should probably do some encapsulation of the runner to reduce potential problems

ratelimit_test.go
rabbbit: Hm.

Agree that having a smaller surface is a good thing, but not sure if this won't over-complicate things here. Like,
- there are literally two tests (one of which tests nothing, I think)
- we're unlikely to have more tests
- if we cause a deadlock than it's on us ;)

Maybe I should have called the `runner` just `args` - would have been closer to my goal of just reducing the noise.

If I pass in the interface I also have to take away access to `r.count`, `r.doneCh` (test below) and I'll need to pass in a function as an argument to `run` (since TestDelayedRateLimiter) is doing some custom work below.

Anyway, can take a stab, why not.

rabbbit: Okay, so if we passed in the interface we kinda need more methods:

```
type runner interface {
	// job tries to Take() on passed in limiter in a loop/goroutine.
	job(rls ...ratelimit.Limiter)
	// waitAssert asserts given count matches after some time.
	waitAssert(t time.Duration, count int)
	// getClock return fake clock.
	getClock() ratelimit.Clock
	// add adds time to the fake clock
	add(t time.Duration)
	// afterFunc executes a func after some time.
	afterFunc(t time.Duration, fn func())
}
```

It's a bit verbose ... but I'm starting to like it - the actual test code looks quite clear.

rabbbit: To be fair, I don't think we'll need `afterFunc` once we fix TestDelayed ... to make sense.

But for now it should stay here.

ratelimit_test.go
```diff
@@ -13,6 +13,82 @@ import (
...
+	maxDuration time.Duration
+	doneCh      chan struct{}
+	wg          sync.WaitGroup
+}
+
+func runTest(t *testing.T, fn func(runner)) {
+	r := runnerImpl{
+		t:      t,
+		clock:  clock.NewMock(),
+		doneCh: make(chan struct{}),
```
cinchurge, line 44: nit: `done` is probably good enough?

ratelimit_test.go
```diff
@@ -13,6 +13,82 @@ import (
...
+	defer close(r.doneCh)
+	defer r.wg.Wait()
+
+	fn(&r)
+	r.clock.Add(r.maxDuration)
+}
+
+// startTaking tries to Take() on passed in limiters in a loop/goroutine.
+func (r *runnerImpl) startTaking(rls ...ratelimit.Limiter) {
+	go func() {
```
cinchurge, line 55: do we need to add this to the waitgroup?

rabbbit: So, technically, I think we should **however** it's not needed right now - `clock.afterFunc` does `time.sleep(1*millisecond)` so we get it for free.

Given that it works now, and the old code was doing this, I'd prefer to keep it as is - just to minimize changes in this PR.

#39 actually adds a separate wg for starting jobs (it's out of date now, I'll update it after this PR) - I'd prefer to iterate & update it in follow ups.

But yeah, more wgs will be needed to deprecate the internal clock for.

ratelimit_test.go
rabbbit: hah, I thought this style was a recommendation? I guess I was wrong :)

cinchurge (approved):
lgtm!

ratelimit_test.go
rabbbit: Oh, sorry, I thought I had pushed/changed this, but apparently didn't.

Will fix in a follow up.

PR #41, opened 2020-11-01
Remove internal fork of andres-erbsen/clock

rabbbit:
Was untested and contained bugs.

This is pending until #39 lands - verified that currently this fails, with #39 landed this passes.


rabbbit:
Merged this into #39.

PR #42, opened 2020-11-01
Provide a way to be able to configure multiple Limiters with same options

rabbbit:
…ions

This is a bit of yack-shaving
- the atomic limiter has a bug - it ignores slack.
- I'd like to add it as a test
- but before doing that, I'd like to have the tests run the same
  on both limiters - to keep them in sync
- to do that, I need be able to configure them in the same way.

Currently. it's a bit of a mess:
- both limiters take Option, but mutexbased simply ignores them
  (because Option only works for the atomic one)

The only way I see to do that stably is to introduce a new stuct
to capture the configuration, and then have each limiter configure
itself based on that struct.

Contrary to the usual setup, having it as a struct should be future
proof - it's an internal struct, so we can change it later on.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -64,21 +64,37 @@ type limiter struct {
 	clock      Clock
 }
 
-// Option configures a Limiter.
-type Option func(l *limiter)
+// config configures a limiter.
+type config struct {
+	maxSlack time.Duration
+	clock    Clock
+}
+
+// buildConfig combines defaults with options.
+func buildConfig(opts []option) config {
+	c := config{
+		maxSlack: 10,
+		clock:    clock.New(),
+	}
+
+	for _, opt := range opts {
+		opt(&c)
+	}
+	return c
+}
+
+// option configures a Limiter.
+type option func(l *config)
 
 // New returns a Limiter that will limit to the given RPS.
-func New(rate int, opts ...Option) Limiter {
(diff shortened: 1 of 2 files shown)


ratelimit.go
```diff
@@ -89,18 +105,16 @@ func New(rate int, opts ...Option) Limiter {
...
 	}
 }
 
 // WithoutSlack is an option for ratelimit.New that initializes the limiter
 // without any initial tolerance for bursts of traffic.
-var WithoutSlack Option = withoutSlackOption
-
-func withoutSlackOption(l *limiter) {
-	l.maxSlack = 0
+var WithoutSlack option = func(c *config) {
```
cinchurge, line 116: why not just provide `WithSlack(slack time.Duration)`? You can easily pass in `WithSlack(0)` while also allowing the caller to set it to any other value.

rabbbit: I do agree this would make more sense.

The reasoning was:
- I tried to keep changes here minimal - keeping the same API
- the WithSlack(5) actually wouldn't work, we just don't respect it.

So I thought I'd fix the slack with actual targeted changes.

cinchurge: i see. agreed that we should keep the same api. we can add new options in a different diff then

cinchurge (approved):

PR #43, opened 2020-12-06
Make tests limiter implementation agnostic 1/2

rabbbit:
The ovarall goal is to be able to run the same tests on all possible
implementations so that we can verify they're working.

After #42 we can now configure limiters with the same options.

After this diff, the tests only pass in those options in to the testRunner.

A following diff will modify testRunner so that all tests are executed
on both atomic & mutex based limiters.

This in turn allows us to change the limiters in sync & unblocks
resolving #27 & #23 (they are about the same bug)

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -14,13 +14,13 @@ import (
 )
 
 type runner interface {
+	// createLimiter builds a limiter with given options.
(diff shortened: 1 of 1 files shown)


rabbbit:
@cinchurge ping ;)

cinchurge (approved):
looks super clean now, thanks!

PR #48, opened 2020-12-07
Push coverage on every push

rabbbit:
Previously we were only pushing coverage on pull requests.
That meant we never had a "base" to compare to, and coverage
report was (1) less usefull (2) complaining about it. See
https://github.com/uber-go/ratelimit/pull/43

This tries to fix it.


cinchurge (approved):
LGTM

PR #49, opened 2020-12-21
Refactor: separate inteface from the implementations

rabbbit:
Just a tiny cleanup - no functional changes:
- keep the "shared/abstract" (config, interface) in ratelimit.go
- move the atomic implementation specific bits into "limiter_atomic"
- move "mutexbased" into "limiter_mutexbased" so that the naming
  is more consistent

Since I'm here I'm also updating the licenses.

The goal is to make reasoning/futher modifications easier, which
changes touching specific implementations only touching specific
implementation files.

### DIFF
--- a/limiter_mutexbased.go
+++ b/limiter_mutexbased.go
@@ -1,4 +1,24 @@
-package ratelimit
+// Copyright (c) 2016,2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ratelimit // import "go.uber.org/ratelimit"
 
 import (
 	"sync"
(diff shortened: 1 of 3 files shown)


ratelimit.go
```diff
@@ -48,28 +45,15 @@ type Clock interface {
...
-}
-
 // config configures a limiter.
 type config struct {
 	maxSlack time.Duration
 	clock    Clock
 }
 
+// New returns a Limiter that will limit to the given RPS.
+var New = newAtomicBased
```
cinchurge, line 55: instead of using a global which can be avoided, can we use a wrapper function instead, and export the separate constructors? this is so that we can return pointers to structs when possible instead of returning interfaces. i.e.
```
func New(...) Limiter  {
  return NewAtomicLimiter(...)
}

func NewAtomicLimiter(...) *AtomicLimiter {...}

func NewMutexLimiter(...) *MutexLimiter {...}
```

rabbbit: Makes sense - done.

There's no immediate benefit of returning (pointers to) structs here though, right? It could make it easier to test some internals of the structs, but that's it - or am I missing something?

cinchurge: Testing is also part of the reason, but it's just good practice in general - by returning an interface instead of a struct, you limit the interface that the exported object is able to implement to just the one you've returned, and it becomes really hard to e.g. add a `.Write()` or `.String()` to the object without modifying the interface.

limiter_mutexbased.go
```diff
@@ -14,8 +34,8 @@ type mutexLimiter struct {
 	clock      Clock
 }
 
-// New returns a Limiter that will limit to the given RPS.
-func newMutexBased(rate int, opts ...Option) Limiter {
+// newMutexBased returns a new atomic based limiter.
+func newMutexBased(rate int, opts ...Option) *mutexLimiter {
```
cinchurge, line 38: I actually think we can export these:
- NewMutexBased
- MutexLimiter
- NewAtomicBased
- AtomicLimiter

returning the struct without exporting them isn't really useful. actually I wonder what @prashantv thinks about this?

rabbbit:
Wait, why?

I don't think we want to expose those structs, and even less the
constructors - we don't intend to expose different implementations, right*?

My understanding was that we're (slowly) moving towards having stable tests
and verified implementations.
Once we have that we will add/benchmark a few more alternatives (mutex
without starvation, ratelimitfx, time/rate), pick the best and stick to
that.

* - we might want to do it if we for whatever reason find that different
implementations work better in some cases, but we're not there yet.

If the concern is breaking changes, we could replace Limiter with a struct
that would wrap/embed the actual implementation.

But I think that's a separate discussion - I'm not changing the API here,
just trying to make the files a bit more organized.

limiter_mutexbased.go
rabbbit: Ops, I responded by email, it didn't attach it here.

Let's copy:

Wait, why?

I don't think we want to expose those structs, and even less the
constructors - we don't intend to expose different implementations, right*?

My understanding was that we're (slowly) moving towards having stable tests
and verified implementations.
Once we have that we will add/benchmark a few more alternatives (mutex
without starvation, ratelimitfx, time/rate), pick the best and stick to
that.

* - we might want to do it if we for whatever reason find that different
implementations work better in some cases, but we're not there yet.

If the concern is breaking changes, we could replace Limiter with a struct
that would wrap/embed the actual implementation.

But I think that's a separate discussion - I'm not changing the API here,
just trying to make the files a bit more organized.

cinchurge: ah ok i see your point. agreed that we don't want to expose atmoic/mutex based implementations to the user, which is too deep in the details for callers. it's unfortunate though that the library started out with returning an interface though :(.

cinchurge (approved):

limiter_mutexbased.go
rabbbit: Meh, we're still in <1.0 - we can change it all ;)

We probably should change it to a struct, but we're not close to 1.0 anyway.

PR #52, opened 2020-12-21
Fix irregularity in benchmark

rabbbit:
The benchmark exetutes the test from 1 to 65k goroutines. Not to take
forever it appropriately uses steps as the number of goroutines grows.

The bucked between 1024 and 2048 has an irregularly low step.
I suspect this is because it's where the atomic limiter becomes
much slower.
However, the small step is not really necessary to show the degredation,
it'll be clearly visible anyway.

Let's make the steps bigger, and make the benchmark run faster.

### DIFF
--- a/ratelimit_bench_test.go
+++ b/ratelimit_bench_test.go
@@ -29,7 +29,7 @@ func BenchmarkRateLimiter(b *testing.B) {
 			for ng := 512; ng < 1024; ng += 32 {
 				runner(b, name, procs, ng, limiter, count)
 			}
-			for ng := 1024; ng < 2048; ng += 8 {
+			for ng := 1024; ng < 2048; ng += 64 {
 				runner(b, name, procs, ng, limiter, count)
 			}
 			for ng := 2048; ng < 4096; ng += 128 {


cinchurge (approved):
i agree it looks irregular, and I am unable to find out why it was written that way. i did accidentally come across #15 which is where the atomic rate limiter was introduced, and it has a pretty cool graph. I'm thinking we should probably add a makefile target to generate that automatically so we can compare across implementations. I can take that on.

rabbbit:
> i did accidentally come across #15 which is where the atomic rate limiter was introduced, 
Yeah, the graphs there are pretty cool.

> i agree it looks irregular, and I am unable to find out why it was written that way.
I suspect this is because where the graph "bends", and we start seeing the difference. But even without the extra detail at that "bend" it it's clear enough.

> I'm thinking we should probably add a makefile target to generate that automatically so we can compare across implementations. I can take that on.
Yeah, graphs would be cool. I was thinking about just `tee | grep | bench_cmp` the current output as a first step, but graphs are always better;)

PR #53, opened 2020-12-21
Make linting fail on warnings

rabbbit:
Currently, the linter just prints a warning if you run the linter,
but returns 0, so linting passed on CI.

After this change we'll start failing for linter warnings,
which is what I believe we want.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ gofmt:
 
 .PHONY: golint
 golint: bin/golint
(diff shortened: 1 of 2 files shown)


rabbbit:
@cinchurge gentle ping. Most of these PRs are tiny, so it shouldn't take too much time.
Alternatively, can/should we move to a post-review model? :)

cinchurge (approved):

PR #21, opened 2019-12-05
Add Per option

kriskowal:
Commits are individually reviewable.
The first changes address nits I missed in review.

The final change addresses #17 by adding an option called Per that reads
ratelimit.New(1, Per(time.Second)) for a 1Hz rate limiter.

PR #54, opened 2021-01-03
Add `Per` option

rabbbit:
Replaces #21

It seems a bunch of people were interested in the feature (see #21),
it makes sense and it's super easy to add.

We also closed #17 in favour of #21, so it's nice to bring it back.


cinchurge (approved):
lgtm!

PR #56, opened 2021-01-03
Run tests on all (two) implementations

rabbbit:
The test runner will simply execute all tests on all implementations.

This is so that we can:
- make sure both implementations are equivalent
- start adding more implementations
- have proper test coverage, yay.

The diff is slightly bigger than the actual changes:
- I'm modifying the tests to run in the same package as the code,
  so that I have access to each of the implementations.
- I'm renaming `runner` to `testRunner` since benchmark defines it's
  own runner.

I considered somehow keeping the `ratelimit_test` package,
and also having a separate `ratelimit` tests to just test
implementations  but that feels too complicated for the benefit - lets
keep it simple and just have a single set of tests.

We could probably also make the benchmark tests re-use the same test runner,
but that's a separate follow up.

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -60,9 +60,10 @@ func runTest(t *testing.T, fn func(testRunner)) {
 	for _, tt := range implementations {
 		t.Run(tt.name, func(t *testing.T) {
 			r := runnerImpl{
-				t:      t,
-				clock:  clock.NewMock(),
-				doneCh: make(chan struct{}),
+				t:           t,
+				clock:       clock.NewMock(),
+				constructor: tt.constructor,
+				doneCh:      make(chan struct{}),
 			}
 			defer close(r.doneCh)
 			defer r.wg.Wait()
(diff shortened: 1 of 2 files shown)


ratelimit_test.go
```diff
@@ -29,36 +28,60 @@ type runner interface {
...
 	wg          sync.WaitGroup
 }
 
-func runTest(t *testing.T, fn func(runner)) {
-	r := runnerImpl{
-		t:      t,
-		clock:  clock.NewMock(),
-		doneCh: make(chan struct{}),
+func runTest(t *testing.T, fn func(testRunner)) {
+	implementations := []struct {
```
abhinav, line 42: minor: This is a bit long. Consider renaming this variable to "tests" to match the usual name for table tests, or "impls" if you think that wouldn't be appropriate since these are not necessarily tests.

ratelimit_test.go
```diff
@@ -111,7 +134,7 @@ func (r *runnerImpl) goWait(fn func()) {
 }
 
 func Example() {
-	rl := ratelimit.New(100) // per second
+	rl := New(100) // per second
```
abhinav, line 137: optional: I like the idea of the example tests showing the full name (`ratelimit.New`) because these show as examples in the documentation. Consider moving them to an `example_test.go` file where we continue to use `package ratelimit_test`.

abhinav (approved):

ratelimit_test.go
rabbbit: Nice, me like.

rabbbit:
thanks @abhinav.

Issue #55, opened 2021-01-03
can you release a version?


AllenX2018:
Hello ~ community maintainer:
the community has not released a version for a long time, can you release a version?

rabbbit:
We haven't released a new version for a long time because not a lot of external functionality has changed in a long time.

We'll release a new 0.X version in a couple of weeks - there's a slack bug we'd like to fix first. In the meantime please use `go get go.uber.com@4fc173c9d3c468c7237a9d7616017378e342bb35`

AllenX2018:
Thank you

rabbbit:
0.2.0 is now released.

PR #60, opened 2021-01-03
Fix not working slack.

rabbbit:
As reported in #23 and #27 slack is currently broken in the atomic
implementation.

Both #23 and #27 propose valid solutions, and I actually arrived
at #23 when trying to solve this independently, but I find #27
slightly more readable/understandable, so going with that.

### DIFF
--- a/limiter_atomic.go
+++ b/limiter_atomic.go
@@ -78,7 +78,6 @@ func (t *atomicLimiter) Take() time.Time {
 			last:     now,
 			sleepFor: oldState.sleepFor,
 		}
-		newState.last = now
 
 		// If this is our first request, then we allow it.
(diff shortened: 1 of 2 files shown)


rabbbit:
ping @cinchurge ;)

Note that #56 is landed now, so we have two implementations to compare against - effectively two tests;)

rabbbit:
@abhinav, could you take a look?

The tests are passing, this is okay even if the atomic implementation still needs more testing & decision.

All this is just so that we don't have obvious bugs & missing features. Once this lands we'll be able to fix a few other slack issues (allow people to configure slack) and maybe release a new minor version (see #55).

cinchurge (approved):
LGTM

PR #61, opened 2021-01-12
add unsafe limiter

mmaxiaolei:
usage scenario: single thread(goroutine)

### DIFF
--- a/limiter_mutexbased.go
+++ b/limiter_mutexbased.go
@@ -34,7 +34,7 @@ type mutexLimiter struct {
 	clock      Clock
 }
 
-// newMutexBased returns a new atomic based limiter.
+// newMutexBased returns a new mutex based limiter.
 func newMutexBased(rate int, opts ...Option) *mutexLimiter {
 	config := buildConfig(opts)
 	l := &mutexLimiter{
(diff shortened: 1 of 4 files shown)


mmaxiaolei:
TRY CLA

rabbbit:
Hi,

Thanks for the contribution!

What's the actual use-case for this? Given the low over-head of the current implementations, and the limitation of a single go-routine, would the "default" limiter not just work?

We thought about possibly extending the API to return different types of rate-limiters depending on the use-case, but I wasn't planning to do it via options. Would need to punt on this a bit more.

mmaxiaolei:
@rabbbit thank you for your reply!

the "default" limiter can work. but we use in cloud native,we want high-performance and low over-head without mutex and cas

and i agree your point about different limiter without options. However, the current design can not be well extended, especially CAS may cause "starvation" in high concurrency scenarios

rabbbit:
> . but we use in cloud native

What do you mean cloud native? Is there a particular application you're needing this in?

> However, the current design can not be well extended, especially CAS may cause "starvation" in high concurrency scenarios

Well, yeah ... but this diff is about having rate limiter in concurrency=1 scenarios, so starvation shouldn't be a problem?
I'm a bit confused now.

rabbbit:
Closing for now - please respond if you still think it's valuable :)

PR #64, opened 2021-02-07
Allow configuring custom slack

### LINKED ISSUE #20: maxSlack can not be self defined.
only WithoutSlack can close the maxSlack. 
What about a new FixedSlack to change default 10?

rabbbit:
Resolves #20

This is to allow people to configure custom slack - we previously
defaulted to 10, and only allowed to disable it.

The only tricky part about this is handling "per" correctly - adding
a bunch of test cases to cover it.

The tests are large, but when I tried to generate them dynamically
(the same test cases 2*, for both with-per and without-per) it became
even harder to read.

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -202,7 +202,6 @@ func TestSlack(t *testing.T) {
 	//   - faster limiter running for 1 second
 	//   - slack accumulated by the faster limiter during the two seconds.
 	//     it was blocked by slower limiter.
-
 	tests := []struct {
 		msg  string
 		opt  []Option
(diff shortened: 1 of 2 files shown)


ratelimit_test.go
```diff
@@ -189,22 +189,91 @@ func TestPer(t *testing.T) {
...
+	// - After 2 seconds, we start another goroutine with
+	//   only the faster limiter. This will allow it to max out,
+	//   and consume all the slack.
+	// - After 3 seconds, we look at the final result, and we expect,
+	//   a sum of:
+	//   - slower limiter running for 3 seconds
+	//   - faster limiter running for 1 second
+	//   - slack accumulated by the faster limiter during the two seconds.
+	//     it was blocked by slower limiter.
+	tests := []struct {
```
rabbbit, line 205: I also had the same test generated dynamically:

```lang=go
	const (
		slowRate = 10
		fastRate = 100
	)

	// Test for a few Per settings.
	testsPer := []time.Duration{
		// Default.
		{time.Second},
		{time.Millisecond * 500},
	}
	// And a few different slack options
	testsSlack := []int{10, 20, 100, 150}

	for _, per := range testsPer {
		for _, slack := range testsSlack {
			testKey := fmt.String("per-%v-slack-%v", per, slack)
			t.Run(testKey, func(t *testing.T) {
				runTest(t, func(r testRunner) {
					slow := r.createLimiter(slowRate, WithoutSlack)
					fast := r.createLimiter(fastRate, tt.opt...)

					r.startTaking(slow, fast)

					r.afterFunc(2, func() {
						r.startTaking(fast)
						r.startTaking(fast)
					})

					// slow limiter with 10hz dominates here - we're always at it's limit.
					r.assertCountAt(1*time.Second, slowRate)

					// results:
					// - for two seconds we ran a slow rate
					// - for a second we ran a fast rate
					// - and we add the slack
					// rates have to be adjusted for Per
					adjustedSlow := slowRate * int(time.Second/testsPer)
					asjustedFast := fastRate * int(time.Second/testsPer)
					r.assertCountAt(3*time.Second, asjustedSlow+adjustedFast+slack)
				})
			})
		}
	}
```

but it seems like manual definition is easier to understand at this point.

cinchurge (approved):
LGTM!

PR #67, opened 2021-02-15
Fix slack ignoring "Per" option

rabbbit:
This is a split off #64 to make #64 smaller/easier.

I believe out current treatment of the default (10) slack doesn't
work very well with Per option.

We configure the limiter with an limit:
- `Limiter(10)` - means, allow 10 requests per second, with 10 requests
  slack - "slack period" is 1 second

However, if we do:
- `Limiter(10, Per(time.Minute))` we get - 10 requests per minute,
  but the slack is still 10 *per second*, meaning the slack period is
  stil 10 second - effectively 1 request.
I believe this is confusing, because with "slack of 10", I'd expect
a 1 minute "slack period".

After this change:
- `Limiter(10, Per(time.Minute))` we get - 10 requests per minute,
  and a slack of 10 requests - so, slack period is 1 minute.

We're basically making it so that both "rate" and "slack" are expressed
in the same unit (ints) and that "Per" applies to both of them.

Note this is currently of minor impact since we're hardcoding the
slack as 10. This will become way more painful/complicated with #64.

### DIFF
--- a/limiter_atomic.go
+++ b/limiter_atomic.go
@@ -45,10 +45,13 @@ type atomicLimiter struct {
 
 // newAtomicBased returns a new atomic based limiter.
 func newAtomicBased(rate int, opts ...Option) *atomicLimiter {
+	// TODO consider moving config building to the implementation
+	// independent code.
 	config := buildConfig(opts)
+	perRequest := config.per / time.Duration(rate)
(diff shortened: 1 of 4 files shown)


cinchurge (approved):
LGTM!

PR #78, opened 2022-03-27
Add link to redit

rabbbit:
People asked about differences. Let's point at the source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,10 @@ func main() {
 }
 ```
 
-See [here][redit] for some historical context.
+## FAQ: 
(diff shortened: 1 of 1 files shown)


abhinav (commented):
Perhaps instead of this, add a minimal FAQ section to the README with just that question, the answer, and the Reddit link?

rabbbit:
> Perhaps instead of this, add a minimal FAQ section to the README with just that question, the answer, and the Reddit link?

Yeah, considered that. Was being lazy. Thanks for the poke.

abhinav (approved):

PR #83, opened 2022-04-30
Correct the comments of Take functions

cncal:
As it can be configured by `Per` function, the `per` no longer always defaults to `time.Second`.

### DIFF
--- a/limiter_atomic.go
+++ b/limiter_atomic.go
@@ -64,7 +64,7 @@ func newAtomicBased(rate int, opts ...Option) *atomicLimiter {
 }
 
 // Take blocks to ensure that the time spent between multiple
-// Take calls is on average time.Second/rate.
+// Take calls is on average per/rate.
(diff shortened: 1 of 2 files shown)


cncal:
https://github.com/uber-go/ratelimit/blob/bf0eb9e5d6347f3021e739b7a77b0ab40916fefa/limiter_mutexbased.go#L37

It should be `a new mutex based limiter` ? If so, I can correct it in this PR.

rabbbit:
Look good, thanks for the correction, sorry for the delay.

> https://github.com/uber-go/ratelimit/blob/bf0eb9e5d6347f3021e739b7a77b0ab40916fefa/limiter_mutexbased.go#L37
> 
> It should be `a new mutex based limiter` ? If so, I can correct it in this PR.

No, this would be against Go docstring guideliness - for function `X` the comments should be in the format `X does` or `X is`, etc.

PR #86, opened 2022-06-05
Setup benchmarks stats and graphs

storozhukBM:
For more details look at: https://github.com/uber-go/ratelimit/pull/85#discussion_r889650431


storozhukBM:
@rabbbit 
Hmmm, the atomic limiter test has failed even though I didn't make any changes

tools/go.mod
```diff
@@ -4,8 +4,10 @@ go 1.14
```
rabbbit, line 4: I'll bump the go version here to get rid of the transient dependencies.

rabbbit:
> @abhinav @rabbbit Hmmm, the atomic limiter test has failed even though I didn't make any changes

Yeah, we've seen this happen (year+ ago, especially during codegen generation), but didn't have time to debug. There's some unknown flakiness, maybe this is a good push for me to actually take a look :)

rabbbit:
Thanks!

rabbbit:
Hey @storozhukBM I'm getting 

image

when running benchmarks on main. I didn't look into this, I assumed you'd know what's going on :)

storozhukBM:
@rabbbit yes, `benchart` requires go 1.18 in its go mod, if you want I can lower to 1.17, but it is better to run benchmarks on latest Go version

rabbbit:
> @rabbbit yes, `benchart` requires go 1.18 in its go mod, if you want I can lower to 1.17, but it is better to run benchmarks on latest Go version

Hm. Technically we want to support two latest versions .. but I guess for benchmarks it's okay. Trying to fix with https://github.com/uber-go/ratelimit/pull/88 - not sure how to test actions without landing them first 🤷

storozhukBM:
I'm not sure what the point of running benchmarks on Github actions is since they run on some virtual machines with unpredictable performance that can vary from run to run.

rabbbit:
It's executed only when new commit is pushed on main, which is probably
around once a year on average :)

I won't pay too much attention to it, but I thought it's something "nice to
have" to perhaps take a look and see in the UI. Also, it tests whether the
benchmarks are runnable.

PR #88, opened 2022-06-05
Only benchmark on the latest Go version

rabbbit:
See  https://github.com/uber-go/ratelimit/issues/86#issuecomment-1146895304

Also:
- update to checkoutv3 (why not)
- use setup-go action to both:
  - pick up the lastest go version
  - do appropriate caching

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -10,22 +10,10 @@ jobs:
     name: Benchmark
     runs-on: ubuntu-latest
     steps:
-      - name: Setup Go
-        uses: actions/setup-go@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
         with:
-          go-version: ${{ matrix.go }}
-
-      - name: Checkout code
-        uses: actions/checkout@v2
-
-      - name: Load cached dependencies
-        uses: actions/cache@v2
-        with:
-          path: ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -10,22 +10,10 @@ jobs:
     name: Benchmark
     runs-on: ubuntu-latest
     steps:
-      - name: Setup Go
-        uses: actions/setup-go@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
```
rabbbit, line 14: I'm switching order here based on https://github.com/actions/setup-go - they seem to recommend first, then setup-go.

I guess it makes sense given we want `setup-go` to do all the caching too.

storozhukBM:
It failed after merge because it still runs on `go version go1.17.10 linux/amd64`

storozhukBM:
Easy way to fix it would be to move
```
	@$(GOBIN)/benchstat stat.txt
	@$(GOBIN)/benchstat -csv stat.txt > stat.csv
	@$(GOBIN)/benchart 'RateLimiter;xAxisType=log' stat.csv stat.html
	@open stat.html
```

into a separate `make` target

rabbbit:
Yeah, I was expecting the `latest` in the action to execute Go1.18. Somehow it didn't - I deferred debuging the action.

Separately make target sgtm too.

PR #89, opened 2022-06-06
Use actively maintained clock library

storozhukBM:
It looks like our clock mocking library is supper old and archived by author.
 It can be a cause of our test instabilities.
 This PR replaces it with drop-in replacement that is actively maintained.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module go.uber.org/ratelimit
 go 1.18
 
 require (
-	github.com/andres-erbsen/clock v0.0.0-20160526145045-9e14626cd129
+	github.com/benbjohnson/clock v1.3.0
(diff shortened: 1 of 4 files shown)


rabbbit:
There's a bit of a larger history here.

We used to have an internal fork of the andres-erbsen clock to work around a bug there. However, our own fork ended up having two bugs (not sure if they were present in anders-erbsen). See #39.

At the time I looked at history between anders-erbsen and the clock you're proposing, both had commits that were not being present on the other one, so I didn't move back to the other fork. And the bobjohnson one had a gap in pushes before the 2021 push by djmitche.

That being said, at this point, you're right that it probably makes sense to just use the "more actively" and hope we don't hit more bugs :]

Issue #90, opened 2022-06-09
atomicInt64Limiter WithoutSlack doesn't block


twelsh-aw:
First off thanks for the library :)

Saw that a new rate limiter was introduced that benchmarked a lot better and pulled it down to try it out.

Noticed that when running WithoutSlack, it just allows everything through instead of waiting because all subsequent Take() calls fall into the `case now-timeOfNextPermissionIssue > int64(t.maxSlack)`

Easiest way to repro is using your `example_test.go`:

- Using `rl := ratelimit.New(100)` (slack=10):
```shell
go test -run Example -count=1
=== RUN   Example
--- PASS: Example (0.09s)
PASS
ok      command-line-arguments  0.207s
```

- Using `rl := ratelimit.New(100, ratelimit.WithoutSlack)`
```shell
go test -run Example -count=1   
--- FAIL: Example (0.01s)
got:
1 10ms
2 775µs
3 3µs
4 2µs
5 10µs
6 2µs
7 2µs
8 2µs
9 2µs
want:
1 10ms
2 10ms
3 10ms
4 10ms
5 10ms
6 10ms
7 10ms
8 10ms
9 10ms
FAIL
FAIL    command-line-arguments  0.126s
FAIL
```

- Using 0.2.0 `rl :=newAtomicBased(100, WithoutSlack)`
```shell
go test -run Example -count=1
PASS
ok      go.uber.org/ratelimit   0.323s
```

I am not 100% sure why the other units with mocked clocks are passing, but your example test and my application tests fail consistently with this new limiter. On darwin if that helps.

storozhukBM:
@twelsh-aw 
Yes, the issue is definitely there, I have a quick fix for it, but I also want to spend a bit more time to figure out why our tests don't catch that issue.

storozhukBM:
@rabbbit
I figured out that our tests continue to pass even if I break currently existing implementation.
For example if you change `ratelimit_test.go:108` to `t.clock.Sleep(interval-interval)` tests will still pass.

storozhukBM:
@rabbbit
Further findings. Tests do not able to detect issues because with mock clock, literally time, freezes between `Take()` calls, to when we check `now.Sub(oldState.last)` in `atomicLimiter` or when we check `now-timeOfNextPermissionIssue` in `atomicInt64Limiter` it always returns 0

storozhukBM:
@rabbbit 
[...]

PR #93, opened 2022-06-27
Fix test approach for detecting issues

storozhukBM:
The problem in the previous approach was that mock time literally didn't run between `limite.Take()` calls, so `now.Sub(oldState.last)` would always be 0 that can be problematic for some issues detection.

This PR adds time increments between `limite.Take()` calls in `(r *runnerImpl) startTaking`. Unfortunately our library for clock mocking `github.com/benbjohnson/clock` has some races reported here https://github.com/benbjohnson/clock/issues/44 and here https://github.com/benbjohnson/clock/pull/42.

So I decided to use the time mocking approach that Ian Lance Taylor used 19 days ago in https://github.com/golang/time repository here https://github.com/golang/time/commit/579cf78fd858857c0d766e0d63eb2b0ccf29f436

Note: this tests approach detects an error in int64 based limiter implementation that the previous test didn't manage to catch.


ratelimit_test.go
```diff
@@ -7,10 +7,105 @@ import (
 
 	"go.uber.org/atomic"
 
-	"github.com/benbjohnson/clock"
 	"github.com/stretchr/testify/assert"
 )
 
+// testTime is a fake time used for testing.
+type testTime struct {
```
rabbbit, line 14: If this is copied over from https://github.com/golang/time/commit/579cf78fd858857c0d766e0d63eb2b0ccf29f436, could we:
(1) put this is a separate file
(2) attribute the source clearly
(3) call out differences in the implementation? For example, I see that `advanceUnlocked` is different than the reference implementation, but I cannot easily tell why.

rabbbit:
Thanks!

This looks complicated, it'll take me some time to grok this (sorry) - I would expect maybe some time over the weekend.

It's interesting that https://github.com/benbjohnson/clock/pull/42 was fixed by @abhinav, looks like the clock bit us a few times.

> Note: this tests approach detects an error in int64 based limiter implementation that the previous test didn't manage to catch.

@storozhukBM - would you like to partially revert #91?  We could add your new implementation to the repo, but not make it the default (yet). This way we could track the (1) clock change (2) bug fix in the same PR eventually. I can also do this, but I thought you might want the addition to be attributed to you in `git log` :)

ratelimit_test.go
storozhukBM: done

storozhukBM:
> @storozhukBM - would you like to partially revert https://github.com/uber-go/ratelimit/pull/91? We could add your new implementation to the repo, but not make it the default (yet). This way we could track the (1) clock change (2) bug fix in the same PR eventually. I can also do this, but I thought you might want the addition to be attributed to you in git log :)

2 Questions:
Should I do the revert in a separate PR or this one?
Should I fix the but in the int64 based implementation right away, since I already have a fix?

rabbbit:
(1) I'd do a separate diff. This one could be about fixing the fake time.
(2) Your call really. I would slightly to push the broken version for now,
and then fix it, but happy either way.

tdlr; is that I'd like to play more with the clock implementations/races
(or at least understand your implementation - which I currently don't)
before landing this PR.

storozhukBM:
@rabbbit 

Let's do the following thing:
1. I'll create a PR with an almost complete revert of https://github.com/uber-go/ratelimit/pull/91, except making int64-based rate limiter a default implementation.
2. After its merge, I'll rebase this branch with a new master, so this branch will start to fail.
3. Then, I'll make a second PR that fixes a known bug.
4. After the fix PR gets merged, I'll rebase this branch again, and tests will show that the bug is fixed.

storozhukBM:
@rabbbit 
First step PR https://github.com/uber-go/ratelimit/pull/94

rabbbit:
Done, #94 merged. sorry for the delay. Thanks!

storozhukBM:
@rabbbit 

Now we can see that new test approach detects issues with `atomic_int64` implementation.

ratelimit_test.go
```diff
@@ -152,6 +169,7 @@ func TestUnlimited(t *testing.T) {
 }
 
 func TestRateLimiter(t *testing.T) {
+	t.Parallel()
```
rabbbit, line 172: Let's do these separately: https://github.com/uber-go/ratelimit/pull/98

storozhukBM: This test approach is slower than the previous and takes 55 seconds without parallelization. That is why I added it to this PR.

rabbbit: Sure, reasonable.

But, since (1) it's reasonable in all cases (2) it will show us the actual slow down of the new approach, let's push it separately?

storozhukBM: Yup, no problem

ratelimit_mock_time_test.go
```diff
@@ -0,0 +1,111 @@
+package ratelimit
+
+/**
+This fake time implementation is a modification of time mocking
+the mechanism used by Ian Lance Taylor in https://github.com/golang/time project
+https://github.com/golang/time/commit/579cf78fd858857c0d766e0d63eb2b0ccf29f436
+
+Modified parts:
+ - advanceUnlocked method uses in-place filtering of timers
```
rabbbit, line 9: is the goal of in-place filtering just performance of the fake-clock?

I don't think this feels very important (?) - I'd rather have simpler code that's slightly slower?

ratelimit_mock_time_test.go
```diff
@@ -0,0 +1,111 @@
...
+			if i != j {
+				tt.timers[j] = tt.timers[i]
+			}
+			i++
+			j++
+		} else {
+			tt.timers[i].ch <- tt.cur
+			for i := 0; i < 16; i++ {
+				runtime.Gosched()
+			}
```
rabbbit, line 90: This also not present in https://github.com/golang/time/commit/579cf78fd858857c0d766e0d63eb2b0ccf29f436.

I guess you're saying we need to gosched 16 times to get the desired consistency?

ratelimit_test.go
```diff
@@ -7,10 +7,18 @@ import (
 
 	"go.uber.org/atomic"
 
-	"github.com/benbjohnson/clock"
 	"github.com/stretchr/testify/assert"
 )
 
+func (tt *testTime) Now() time.Time {
```
rabbbit, line 13: why did you keep it here as opposed to the other file with other methods?

ratelimit_test.go
```diff
@@ -64,21 +72,28 @@ func runTest(t *testing.T, fn func(testRunner)) {
...
+				move := func() {
+					defer func() {
+						_ = recover()
+						time.Sleep(10 * time.Millisecond)
+					}()
+					r.clock.advanceToTimer()
+				}
+				for {
+					move()
+				}
```
rabbbit, line 95: Sorry, but I think this needs a much larger explanation - perhaps as comments in code?

Am I reading this right that this is an infinite loop, trying to move to infinitely move to the next timer ... by recovering from a panic?

Irrespective of goroutine leaks, using panics for signaling - do you think the `maxDuration` is not useful here? It seems we should be able to just pass the duration into `advance`, theoretically?

ratelimit_mock_time_test.go
storozhukBM: For me, this approach looks simpler as well as more performant, but check the original and give me your opinion, if you think we should keep the original, I don't mind.

ratelimit_mock_time_test.go
storozhukBM: We just run much more goroutines concurrently than golang/time tests, so we need to give them more time to run their logic.

ratelimit_test.go
storozhukBM: I thought that we could keep in ratelimit_mock_time_test.go things that were adopted from golang/time and here we just implement our own clock interface.

ratelimit_test.go
storozhukBM: > Irrespective of goroutine leaks

I think 1 goroutine leak per test case is not a big deal, but if you want, I can add some exit logic here.

>  using panics for signaling

Code from Ian just panics when you try to run the next timer, and there is nothing to run. We can potentially change that behavior to not handle panics, but I wanted to limit changes in that code.

> Sorry, but I think this needs a much larger explanation - perhaps as comments in code?

I'll write some description here first and when we clarify all questions about it, I'll move it into code comments:

```
testTime internally has a queue of all timers currently registered within the clock.
In our case, there will be timers for goroutines that try to consume permissions and timers 
for goroutines that should run test asserts. When we call testTime.advanceToTimer() it basically
shifts time to the next registered timer, runs it, and also runs runtime.Gosched() multiple times 
to give a chance for other goroutines to run and react to what happened in the current timer. 
Usually, when we run runtime.Gosched() other goroutines will observe the new state, make their moves, and potentially create new timers inside this clock to wait for some other state change. 

If testTime.advanceToTimer() panics, it means that there are no timers to run, 
and it can happen if the test is alre
```

 > Do you think the maxDuration is not useful here?

No, we can use that if we want to get out of the currently infinite loop, we can check on every iteration if testTime is already passed maxDuration.

>  It seems we should be able to just pass the duration into advance, theoretically?

Yes, but since we advance time in different goroutines in this approach, it is necessary for us to do it one step at a time and give other goroutines the opportunity to make their steps as well. There also can be situations where we start to advance time in this goroutine before test goroutines spawned all permission consumers and asserting goroutines, so we need to just wait for everything to setup, that is what we do when panic happens.

ratelimit_mock_time_test.go
rabbbit: Weird, I wish there was a way to tell runtime "run everything else but me". And `time.Sleep(Nanosecond) is not great.

Ack.

ratelimit_test.go
rabbbit: Reasonable idea.

Separately, I'm surprised we don't get a linter failures about missing docstrings. Some checks are missing, but that's not related to your change.

ratelimit_mock_time_test.go
rabbbit: It's just that the whole `i`, `j` looks like we would need tests for tests (off by one errors).

I think I would rather pull the other version as is, and then push a separate PR to fix the copying with clear changes. If we worried about the performance though, wouldn't we want to sort once on insert, rather than `advance`?

But:
- see my other comment below
- I understand if fixing the time is not your primary focus for now :)

ratelimit_test.go
rabbbit: So perhaps naively, but do we need both
- advanceToTimer
- advance(Unlocked?)

Our test runner knows exactly how much time needs to move (`maxDuration`). We can just pass it into `advance`. `advanceUnlocked` already does all the `runtime.Gosched` needed.

And soit seem to me that `advance` should theoretically be enough?
If the problem is the initial goroutine setup, should we just add the `16*runtime.Gosched` at the **beginning** of `advance` to let the scheduler kick in?
This would also help with `r.clock.advance(time.Nanosecond)` you have below.

I understand this kinda modifies the code we copy-paste, but I'd feel much more comfortable with we decreased the API by 50% here.

ratelimit_mock_time_test.go
storozhukBM: OK, I can go back to the prev version.

ratelimit_test.go
storozhukBM: @rabbbit

Right now, it is hard for me to make changes because tests do not pass until we merge https://github.com/uber-go/ratelimit/pull/95

I agree with the comments and changes that you propose right now, but they don't change the approach in general. 
I propose to merge https://github.com/uber-go/ratelimit/pull/95, and then I can apply your suggested changes, or we can merge this PR as it is right now, and I'll make these improvements in subsequent PRs.

rabbbit: Merged #95. I might still want to re-review it later on, it feels like some simplification is possible.

storozhukBM: I'll update this branch with new version and this should potentially fix all tests, event though I'm not sure how your new test for initial calls will work 🤞

storozhukBM:
@rabbbit 
After merging with the recent main branch, all tests pass, except TestInitial for mutex rate limiter for some reason. I'll investigate what happens with it when I have time in the next few days.

rabbbit:
It's possible that mutex behaviour was broken, but we didn't detect it
because of the broken clock :|

storozhukBM:
@rabbbit 

> It's possible that mutex behaviour was broken, but we didn't detect it
because of the broken clock :|

No, the use of `advanceToTimer` approach removes the issue. I think just `advance(time.Duration)` is not sufficient with this `testTime` implementation. I'll figure it out.

storozhukBM:
@rabbbit 
I think I addressed all your comments, so please take another look

ratelimit_test.go
```diff
@@ -7,11 +7,22 @@ import (
...
+	return tt.now()
+}
+
+func (tt *testTime) Sleep(duration time.Duration) {
+	timer, _ := tt.newTimer(duration)
+	<-timer
+}
+
 type testRunner interface {
+	getName() string
```
rabbbit, line 25: why do we need this?

The name is included in the test-name automatically via tt.run (?)

ratelimit_test.go
```diff
@@ -7,11 +7,22 @@ import (
 
 	"go.uber.org/atomic"
 
-	"github.com/benbjohnson/clock"
 	"github.com/stretchr/testify/assert"
 )
 
+const advanceBackoffDuration = 5 * time.Millisecond
```
rabbbit, line 13: Do you know why we need this? Is the `gosched` not enough?

ratelimit_test.go
```diff
@@ -102,6 +115,7 @@ func (r *runnerImpl) startTaking(rls ...Limiter) {
 			for _, rl := range rls {
 				rl.Take()
 			}
+			r.clock.advance(time.Nanosecond, 0)
```
rabbbit, line 118: and we need this one as well?

rabbbit (approved):
Almost SGTM:
- I don't think we need the `GetName` (?)
- I don't like the amount of `time.Sleep` but perhaps it's unavoidable 

I've been wondering about some simplified fake clock library - given the amount of issues we've seen it seems it must be possible to make it simpler -__-

ratelimit_test.go
storozhukBM: agree

ratelimit_test.go
storozhukBM: You can notice that `gosched` calls happen under lock. It means that all goroutines that would like to get the current time from this clock or register some additional timer won't be able to do so, even if we run these gosched calls. In our case, it means that most "startTaking" goroutines will never work while we step through timers, but they need the most time resource in our tests run.

So we actually need both types of pauses between runs. One is using `gosched` under lock for all goroutines that were waiting for the timer to click and are not blocked on mutex of this clock. 
The other one is for all goroutines that are blocked on mutex of this clock and need to make some progress between timer advances.

ratelimit_test.go
storozhukBM: Yes, this one imitates time progressions between `rl.Take()` calls. With a real clock, when you call `rl.Take()` do something and call `rl.Take()` again `time.Now()` will move forward in time, so we want to imitate that.

storozhukBM:
> I've been wondering about some simplified fake clock library - given the amount of issues we've seen it seems it must be possible to make it simpler -__-

If you take a look at the libraries we have used in the past, like https://github.com/benbjohnson/clock/ it is MUCH more complex, has concurrency bugs, and `time. Sleep's as well [here](https://github.com/benbjohnson/clock/blob/68df829297d4f02d5997a02f8c0e8d4b12f0b2a3/clock.go#L373)

I researched every library that I was able to find on Github, and this approach from `golang/time` is the most straightforward I've seen.

One more consideration that we should take into account is that we can find a lib or implementation that will pass all our tests but is going to detect the real bug. After we change the time library, we should use [mutation testing](https://en.wikipedia.org/wiki/Mutation_testing) at least manually introduce some bugs into the existing rate limiter implementations that we have and make sure that our tests with the new clock library detect the issue.

storozhukBM:
Hey @rabbbit, what do you think? Should we move forward with this PR?

rabbbit:
We spoke about this briefly in #101 and #90 - I'm not sure about merging this:
- without in an depth review
- without a plan for this library

For now I'll put this as "on-hold". This means that the library is effectively in code-freeze, as our test clock implementation is buggy.

storozhukBM:
@rabbbit  can we find someone form Uber to make a depth review?

storozhukBM:
We maybe can get something in standard library for this https://github.com/golang/go/issues/67434

PR #94, opened 2022-06-29
Restore int64 based atomic rate limiter

storozhukBM:
This limiter was introduced and merged in the following PR [#85](https://github.com/uber-go/ratelimit/pull/85)
Later [@twelsh-aw](https://github.com/twelsh-aw) found an issue with this implementation [#90](https://github.com/uber-go/ratelimit/issues/90)
So [@rabbbit](https://github.com/rabbbit) reverted this change in [#91](https://github.com/uber-go/ratelimit/pull/91)

Our tests did not detect this issue, so we have a separate PR [#93](https://github.com/uber-go/ratelimit/pull/93) that enhances our tests approach to detect potential errors better.
With this PR, we want to restore the int64-based atomic rate limiter implementation as a non-default rate limiter and then check that [#93](https://github.com/uber-go/ratelimit/pull/93) will detect the bug.
Right after it, I'll open a subsequent PR to fix this bug.

### DIFF
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -11,6 +11,7 @@ require (
 
 require (
 	github.com/BurntSushi/toml v1.0.0 // indirect
+	github.com/storozhukBM/benchart v1.0.0
 	golang.org/x/exp/typeparams v0.0.0-20220328175248-053ad81199eb // indirect
 	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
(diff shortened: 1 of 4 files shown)


tools/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 
 require (
 	github.com/BurntSushi/toml v1.0.0 // indirect
+	github.com/storozhukBM/benchart v1.0.0
```
rabbbit, line 14: This one looks like we missed it earlier somewhere.

rabbbit (approved):

rabbbit:
Thanks!

PR #95, opened 2022-07-02
Fix no slack option for int64 based option

storozhukBM:
This PR fixes the issue found by [@twelsh-aw](https://github.com/twelsh-aw) with int64 based implementation https://github.com/uber-go/ratelimit/issues/90

Our tests did not detect this issue, so we have a separate PR https://github.com/uber-go/ratelimit/pull/93 that enhances our tests approach to detect potential errors better.

### DIFF
--- a/limiter_atomic_int64.go
+++ b/limiter_atomic_int64.go
@@ -66,10 +66,10 @@ func (t *atomicInt64Limiter) Take() time.Time {
 		timeOfNextPermissionIssue := atomic.LoadInt64(&t.state)
 
 		switch {
-		case timeOfNextPermissionIssue == 0:
-			// If this is our first request, then we allow it.
+		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
+			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
 			newTimeOfNextPermissionIssue = now
-		case now-timeOfNextPermissionIssue > int64(t.maxSlack):
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack):
 			// a lot of nanoseconds passed since the last Take call
 			// we will limit max accumulated time to maxSlack
 			newTimeOfNextPermissionIssue = now - int64(t.maxSlack)
@@ -82,9 +82,6 @@ func (t *atomicInt64Limiter) Take() time.Time {
 			break
 		}
 	}
-	nanosToSleepUntilOurPermissionIsIssued := newTimeOfNextPermissionIssue - now
-	if nanosToSleepUntilOurPermissionIsIssued > 0 {
-		t.clock.Sleep(time.Duration(nanosToSleepUntilOurPermissionIsIssued))
-	}
+	t.clock.Sleep(time.Duration(newTimeOfNextPermissionIssue - now))
 	return time.Unix(0, newTimeOfNextPermissionIssue)
 }


storozhukBM:
No perf degradation detected due to this fix

IMAGE 2022-07-02 22:01:03

limiter_atomic_int64.go
```diff
@@ -66,10 +66,10 @@ func (t *atomicInt64Limiter) Take() time.Time {
 		timeOfNextPermissionIssue := atomic.LoadInt64(&t.state)
 
 		switch {
-		case timeOfNextPermissionIssue == 0:
-			// If this is our first request, then we allow it.
+		case t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest):
```
rabbbit, line 69: I still need to grok the code, but briefly looking at this, I wonder if we should do:

```golang
if t.maxSlack == 0 {
  t.maxSlack == t.preRequest
}
```
during the initialization to get rid of one of the cases.

rabbbit (commented):
I'm really sorry about relaying this.

This is pretty top on my priority list I'll really try to get this soon.

(will need to understand the testing/time changes too)

limiter_atomic_int64.go
storozhukBM: Unfortunately, no, because this will allow for one additional permission to accumulate.
In general, I think we can simplify this and other implementations if we change the way how we initialize rate limiter state, but this is a subject for a separate PR, I think.

rabbbit: Hm. I will need to really think about this more.

In your current proposal though, are we not changing the behaviour for the very first request?

I might add tests for the first request only to formalize the behavior.

rabbbit: I'm less worried about implementation simplification than any unspecified behavior - I'd like to make sure that all implementation behave identically.

Tried to address initial few "Takes" in #97 - WDYT?

I might add one or two more tests before looking at the "time" changes.

storozhukBM: > I'm less worried about implementation simplification than any unspecified behavior - I'd like to make sure that all implementation behave identically.

This fix functional behaves identically to the previous behavior. It just applies the same behavior to cases when no one has used RL for a long time, so it accumulated many permissions. But this limiter has no slack, so we should not allow permissions to accumulate, so we need to move the `newTimeOfNextPermissionIssue` to `now`, as if RL was created right now.

storozhukBM:
@rabbbit 
New test for initial call failed, but only on 1.18, bad thing is that it doesn't show what exact limiter type didn't pass the test. 
Maybe we should add additional test logging for that.

storozhukBM:
@rabbbit 
I did the same force-push 😄 already

limiter_atomic_int64.go
```diff
@@ -66,10 +66,10 @@ func (t *atomicInt64Limiter) Take() time.Time {
 		timeOfNextPermissionIssue := atomic.LoadInt64(&t.state)
 
 		switch {
-		case timeOfNextPermissionIssue == 0:
-			// If this is our first request, then we allow it.
+		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
+			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
 			newTimeOfNextPermissionIssue = now
-		case now-timeOfNextPermissionIssue > int64(t.maxSlack):
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack):
```
rabbbit, line 72:
```suggestion
		now-timeOfNextPermissionIssue > int64(t.maxSlack):
```
I think the maxSlack check here is unnecessary - the previous iff should have covered that.

rabbbit (approved):

rabbbit:
Since this is not actually the "main" limiter yet, I'll merge this in and potentially post-review later on.

limiter_atomic_int64.go
storozhukBM: I think there can be a case where `t.maxSlack == 0` is true, but `now-timeOfNextPermissionIssue > int64(t.perRequest)` is false, in that case we will try to evaluate `now-timeOfNextPermissionIssue > int64(t.maxSlack)` and it can be true, so we will end up in a wrong branch. So we need `t.maxSlack > 0` check here to step into `default` branch.

PR #97, opened 2022-07-10
Add a test verifying initial startup sequence

rabbbit:
See https://github.com/uber-go/ratelimit/pull/95#discussion_r915251700

From that discussion I wasn't sure whether the proposed the initial
startup sequence of the limiter - i.e. whether at startup we always
block, or always allow.

Since we didn't seem to have that codified (perhaps apart from the
`example_test.go`) this PR adds a test to verify this.

This is still slightly (2/1000) flaky, but I think that's good enough
to add this in - should be valuable anyway.

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -226,29 +226,30 @@ func TestInitial(t *testing.T) {
 				rl := r.createLimiter(10, tt.opts...)
 
 				var (
-					clk            = r.getClock()
-					prev time.Time = clk.Now()
+					clk  = r.getClock()
+					prev = clk.Now()
 
-					results         []time.Duration
-					startWg, doneWg sync.WaitGroup
+					results = make(chan time.Time)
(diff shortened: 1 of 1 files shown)


rabbbit:
Well this one clearly needs more work.

Works fine 998/1000 on my machine, but github runner is consistently unhappy.

rabbbit:
> Well this one clearly needs more work.
> 
> Works fine 998/1000 on my machine, but github runner is consistently unhappy.

Turns out I just wasn't running `-race` locally. This is interesting though - because we have RL the code should not be racy, but the race detector (I think) has no way of knowing this.

rabbbit:
@storozhukBM Could you take a look?

Irrespective of the implementation changes I think we have a gap WRT behavior on the initial few requests. This seems useful to formalize in tests.

storozhukBM (approved):
LGTM

PR #98, opened 2022-07-10
Run tests in parallel.

rabbbit:
In #93 @storozhukBM adds test parallization, which is unrelated
to the actual change. I'd rather have this as a separate PR.

The test currently take ~12 seconds on my laptop, so I see why we might
want to run them in parallel. There are other parallelization
opportunities (we could paralelize each of the subtests for each
of the rate-limiters), but this is a good start.

Before:
```
> go test ./... -race -count 1
ok      go.uber.org/ratelimit   12.739s
```

After:
```
> go test ./... -race -count 1
ok      go.uber.org/ratelimit   6.178s
```

### DIFF
--- a/ratelimit_test.go
+++ b/ratelimit_test.go
@@ -143,6 +143,7 @@ func (r *runnerImpl) goWait(fn func()) {
 }
 
 func TestUnlimited(t *testing.T) {
+	t.Parallel()
 	now := time.Now()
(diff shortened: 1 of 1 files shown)


rabbbit:
Given this is straighforward, should have no impact and only affects development, merging this in.

PR #101, opened 2022-10-31
Make int64 based atomic ratelimiter default

storozhukBM:


limiter_atomic_int64.go
```diff
@@ -82,6 +82,12 @@ func (t *atomicInt64Limiter) Take() time.Time {
...
 		}
 	}
-	t.clock.Sleep(time.Duration(newTimeOfNextPermissionIssue - now))
-	return time.Unix(0, newTimeOfNextPermissionIssue)
+
+	sleepDuration := time.Duration(newTimeOfNextPermissionIssue - now)
+	if sleepDuration > 0 {
+		t.clock.Sleep(sleepDuration)
+		return time.Unix(0, newTimeOfNextPermissionIssue)
+	}
```
rabbbit, line 90: Okay, now I'm confused again :)

You removed very similar lines in #95 (https://github.com/uber-go/ratelimit/pull/95/files#diff-30af80b2f0055e05bdc7d6ecea1b49d55f1ef946c381a1604f8e03d58789cd3fL85) and #93 which "fixes the time" doesn't have this change.

Why re-introduce this now?

storozhukBM: 4 month ago, you added a new test `Example_default` that checks the slack behaviour of the default ratelimiter.
It turns out, `atomicLimiter` returns `now` if it doesn't sleep, so I decided to sync this behaviour in `atomicInt64Limiter`.

> you removed very similar lines in https://github.com/uber-go/ratelimit/pull/95

Yes, these lines are similar, but slightly different in the current PR we return `now` if we don't sleep, and this is the only behaviour change, everything else works as it worked before.

> https://github.com/uber-go/ratelimit/pull/93 which "fixes the time" - which I assumed is the goal-state - doesn't have this change.

In #93 `atomicLimiter` is still a default implementation, so I didn't know about the difference between behaviours so far.

storozhukBM: @rabbbit ^^

rabbbit (approved):

PR #104, opened 2023-01-16
Clarify installation

matthiasgubler:


matthiasgubler:
Ping @abhinav for review :)

rabbbit:
Sorry, I don't think this adds enough value - this library is no different than any other Go module, so users can just follow whatever they do typically.

PR #116, opened 2023-07-09
Update clock dependency

rabbbit:
There were quite a few fixes in that repo, let's pull them in.

We've had a few mock-clock bugs before (#90), I'm hoping that with the new versions #93 won't be necessary. I'll try reverting #95 on a branch temporarily to see if it would have been caught.


rabbbit:
This one didn't work very well. I *might* look at this in detail, but otherwise we're frozen until this is resolved.

Issue #119, opened 2023-12-01
Take always return without block if it enters the case branch


smallnest:
https://github.com/uber-go/ratelimit/blob/main/limiter_atomic_int64.go#L72
```go
		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack):
			// a lot of nanoseconds passed since the last Take call
			// we will limit max accumulated time to maxSlack
			newTimeOfNextPermissionIssue = now - int64(t.maxSlack)
```
 you will find the case always (almost) returns true.

The below code is a test for reproduce:
```go
func TestLimiter(t *testing.T) {
	limiter := ratelimit.New(1, ratelimit.Per(time.Second), ratelimit.WithSlack(1))

	for i := 0; i < 25; i++ {
		if i == 1 {
			time.Sleep(2 * time.Second)
		}
		limiter.Take()

		fmt.Println(time.Now().Unix(), i) // burst

	}
}
```

ratelimit is v0.3.0

storozhukBM:
@smallnest can you please try the version from #120 and confirm that it works for you

smallnest:
yes. Fixed. thanks

spencercjh:
The fixed PR isn't merged, please re-open the issue to noticed the maintainer and other users to pay attention to the problem.

Issue #106, opened 2023-06-21
When one of my operations has not called the atomicInt64Limiter.task() function for a long time, the speed limit may fail to be called again


zoutao0909:
Here is my test function：
func TestName(t *testing.T) {
	rl := New(1)

	prev := time.Now()

	var sizes = 0
	for i := 0; i < 100000; i++ {
		sizes++
		if sizes >= 1000 {
			now := rl.Take()
			fmt.Println(sizes, i, now.Sub(prev))
			prev = now
			if 4000 < i && i < 5000 {
				time.Sleep(30 * time.Second)
			}
			sizes = 0
		}
	}
}
the result i expected：
=======================================================================================
1000 999 2µs
1000 1999 1s
1000 2999 1s
1000 3999 1s
1000 4999 1s
1000 5999 30.00147s
1000 6999 1s
1000 7999 1s
1000 8999 1s
1000 9999 1s
1000 10999 1s
1000 11999 1s
1000 12999 1s
=======================================================================================

I get the result：
=======================================================================================
1000 999 3µs
1000 1999 1s
1000 2999 1s
1000 3999 1s
1000 4999 1s
1000 5999 30.001429s
1000 6999 326µs
1000 7999 34µs
1000 8999 7µs
1000 9999 6µs
1000 10999 6µs
1000 11999 6µs
1000 12999 6µs
1000 13999 6µs
1000 14999 6µs
1000 15999 6µs
1000 16999 6µs
1000 17999 6µs
1000 18999 6µs
=======================================================================================

I got the desired result after modifying the code here：

pkg : ratelimit
fileName : limiter_atomic_int64.go
funcName : func (t *atomicInt64Limiter) Take() time.Time
line : 76

update   
newTimeOfNextPermissionIssue = now - int64(t.maxSlack)
to 
newTimeOfNextPermissionIssue = now)

rabbbit:
This is expected - by default the limiter starts with a slack of 10.

Please use `WithoutSlack` option in https://github.com/uber-go/ratelimit/blob/main/ratelimit.go#L101

PR #120, opened 2023-12-01
Fix immediate slack re-apply issue

storozhukBM:
This issue is reported here #119
I was able to reproduce and fix it.

### DIFF
--- a/limiter_atomic_int64.go
+++ b/limiter_atomic_int64.go
@@ -69,7 +69,7 @@ func (t *atomicInt64Limiter) Take() time.Time {
 		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
 			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
 			newTimeOfNextPermissionIssue = now
-		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack):
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack)+int64(t.perRequest):
 			// a lot of nanoseconds passed since the last Take call
 			// we will limit max accumulated time to maxSlack
 			newTimeOfNextPermissionIssue = now - int64(t.maxSlack)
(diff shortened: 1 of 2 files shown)


storozhukBM:
@rabbbit  please take a look

rabbbit:
I'll take a look on Monday.

I'm not a fan of the test, both sleeping and asserting that time passed.

I'm also considering reverting to the basic locking version. This is the second bug we found in this one, so it's clearly beyond both of us to write this confidently.

spencercjh:
This bug causes the failure of the rate limiter when it's a single instance of an online service. For example, a client with intermittent requests to call another service. The client holds a rate limiter as a field. The target service sometimes returns 429 (too many requests). It took me a lot of time to realize that it was a problem in this PR, and I always thought it was something in the way I invoked `Take`. 😅 

PLZ move forward with merging this PR into the master branch ASAP. 🙏 🙏 🙏 

Associated issues: #106, #119.

limiter_atomic_int64.go
```diff
@@ -69,7 +69,7 @@ func (t *atomicInt64Limiter) Take() time.Time {
 		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
 			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
 			newTimeOfNextPermissionIssue = now
-		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack):
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack)+int64(t.perRequest):
```
VovkoO, line 72: Wouldn't this solution make the withSlack and withoutSlack options work exactly the same.
Looks like without slack you condition looks like:
`now - prev > period`
and withSlack
`now - prev - slack > slack + period`
Which is the same thing

storozhukBM: We apply `now - int64(t.maxSlack)` only under condition and that makes the difference.

rabbbit:
Sorry for the delay @storozhukBM - I believe I understand your fix, but propose we do something like https://github.com/uber-go/ratelimit/pull/124 to get the test in-line.

I remember about your clock implementation and I dislike where we are, but I'm not willing to give up on tests totally just yet :)

rabbbit:
Fixed via #124 instead - thank you for the debugging/fix here @storozhukBM.

PR #131, opened 2026-05-15
Update go modules to latest

chimanjain:
Update go modules to latest

* `github.com/benbjohnson/clock`
* `github.com/stretchr/testify`
* `go.uber.org/atomic`

It also fixes CVE - `GO-2022-0603`


chimanjain:
Can the reviewers please verify these changes!

rabbbit:
I do not think we're at risk of the cve. Why are we doing this change?

chimanjain:
> Can the reviewers please verify these changes!

You're completely right. `ratelimit` is not at risk from this CVE since yaml.v3 is only an indirect dependency brought in by testify for tests, and we don't parse untrusted YAML.

The main motivation for this PR is routine dependency maintenance/hygiene. Upgrading testify also has the added benefit of silencing automated vulnerability scanners (govulncheck / Snyk / Dependabot) that flag GO-2022-0603 when scanning go.sum.

I can updated the PR description to clarify that this is a routine dependency bump rather than a critical vulnerability fix.

One more thing, it will help migrate from `gopkg.in/yaml.v3` which has been archived and `go.yaml.in/yaml/v3` is the maintained gomodule by yaml org itself.

Reference: https://github.com/yaml/go-yaml/

